PR #11, opened 2021-05-09
Read light novel fixes

BreadGenie:
- Some novels have the chapter content twice in the ChapterText like

- There's a stray `Sponsored Content` string in the ChapterText like


BreadGenie:
the commit `Fix Chapter content appearing twice in ChapterText` didn't actually fixed it 😓 .
I'll fix it soon.

PR #14, opened 2021-05-12
Add WuxiaWorldSite

BreadGenie:
Source: [WuxiaWorldSite](https://wuxiaworldsite.co/)


src/en/wuxiaworldsite/WuxiaWorldSiteScraper.js
```diff
@@ -0,0 +1,211 @@
...
+
+    novel.Status = novel["Genre(s)"].pop();
+
+    novel["Genre(s)"] = novel["Genre(s)"].join(",");    
+
+    let novelChapters = [];
+
+    $(".new-update-content").each((i, el) => {
+        let chapterName = $(el).text().split(/\t+/);
+        const releaseDate = chapterName.pop();
```
BreadGenie, line 102: `releaseDate` will be shown as xx Years/Months/Days/Hours ago instead of a proper date.
Should I keep it like that or make it null?

src/en/wuxiaworldsite/WuxiaWorldSite.js
```diff
@@ -0,0 +1,14 @@
+const express = require("express");
+const WuxiaWorldScraper = require("./WuxiaWorldSiteScraper");
+
+const router = express.Router();
+
+router.get("/novels/", WuxiaWorldSiteScraper.novelsScraper);
+
+router.get("/novel/:novelID/:novelUrl", WuxiaWorldSiteScraper.novelScraper);
```
rajarsheechatterjee, line 8: Most sources follow the same endpoints for novel and chapter which are `/novel/:novelUrl` and `/novel/:novelUrl/:chapterUrl`  and because of `novelID`  it would not work.  

I checked the website and if we remove the `novelID` from the url the same novel or chapter page still loads so if we just remove the `/:novelID` it would work.

src/en/wuxiaworldsite/WuxiaWorldSiteScraper.js
rajarsheechatterjee: Most sources don't have `releaseDate` so either is fine.

src/en/wuxiaworldsite/WuxiaWorldSite.js
BreadGenie: I should've noticed that. Thanks for pointing it out.

rajarsheechatterjee:
Thanks for your PRs and adding a new source :)

src/en/wuxiaworldsite/WuxiaWorldSite.js
BreadGenie: Fixed it. :D

PR #19, opened 2021-05-16
Try using random user agents for down sources

BreadGenie:
This is just a try to check if it'd work or not. Possibly not mergeable.
I think that it's quite possible that the down sources are using some kind of anti-scraper/DDoS protection service. So I'm using random user agents to try circumventing the DDoS protection.


BreadGenie:
Seems like it worked for MTLNovels but except scraping chapters and search :/

rajarsheechatterjee:
Thanks for fixing this.

I'll merge this and use puppeteer for chapters and search in MTL novel and LightNovelPub.

BreadGenie:
I thought I didn't fix it :0
I missed the part where some of the scraper functions used node fetch. No wonder it worked partially.

Issue #43, opened 2021-07-10
[Source Request] Yushubo


seizue:
Source Name: Yushubo
Link: https://m.yushubo.com/
Language: Chinese

BreadGenie:
Took a look and it seems like they forbid crawlers :/ 
Might need to use puppeteer but it will impact the performance and the chapters will load pretty slow

rajarsheechatterjee:
@BreadGenie I don't think using puppeteer is possible anymore as I've moved the sources to the app.

BreadGenie:
I've managed to make it work using user agent for now.
I'll put together a PR once it's ready

PR #45, opened 2021-07-13
feat(source): Add Yushubo

BreadGenie:
#43 
If this is approved I'll make a PR to the app too


rajarsheechatterjee:
Looks good.

In the PR to the app, change `extensionId` to 51 as 50 is already assigned to another source.

Also, `nextChapter` & `prevChapter` can remain null as they are no longer used in the app.

PR #465, opened 2023-03-15
Plugins host

nyagami:
Many sources were added.

I will leave the rest for someone else xD


libs/htmlToText.js
```diff
@@ -0,0 +1,266 @@
+const htmlToText = (html, options = {}) => {
```
rajarsheechatterjee, line 1: Do we need this now since we only have the WebView reader?

plugins/chinese/.chinese
rajarsheechatterjee, line 1: What are these files?

package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "plugins",
```
rajarsheechatterjee, line 2:
```suggestion
  "name": "lnreader-plugins",
```

plugins/chinese/.chinese
nyagami: just an empty files to make folder visible on github

libs/htmlToText.js
nyagami: Yeah, I will remove it

.env
rajarsheechatterjee, line 1: You could use `.env.example` instead and add this in `.gitignore`.

PR #569, opened 2023-08-21
remove protected attribute

nyagami:
reader doesnt need this anymore


nyagami:
conflits were resolved

PR #639, opened 2023-11-06
plugin refactoring

Rider21:
fixed all Russian plugins
fixed madara tags
removed madara plugins that don't work


nyagami:
There's a conflict because the lastest merged PR modified the `.dist/LNReader`. Pls resolve it.
To avoid conflict, next time, you dont have to edit `.dist/LNReader`. Just leave as it, We already had a github action to generate json for LNReader

Rider21:
Thanks, I hadn't noticed that the github action was enabled.

PR #719, opened 2024-01-24
(fix src) Novel Updates

K1ngfish3r:
this should fix adding filters in NU


plugins/english/nobleMTL.ts
```diff
@@ -152,63 +153,40 @@ class NobleMTL implements Plugin.PluginBase {
         searchTerm: string,
         pageNo?: number | undefined
```
nyagami, line 154:
```suggestion
        pageNo: number

```

plugins/english/novelupdates.ts
```diff
@@ -91,10 +33,58 @@ class NovelUpdates implements Plugin.PluginBase {
...
+        const loadedCheerio = parseHTML(body);
+        return this.parseNovels(loadedCheerio);
+    }
+
     async parseNovelAndChapters(novelUrl: string): Promise<Plugin.SourceNovel> {
         const url = novelUrl;
-
-        const result = await fetchApi(url);
+        const headers = new Headers();
+        const result = await fetchApi(url, { headers });
```
nyagami, line 87:
```suggestion
        const result = await fetchApi(url);
```

nyagami (changes requested):

K1ngfish3r:
oh wait, i need to add in .js as well

nyagami (approved):

PR #725, opened 2024-01-25
Start adding documation and add host.ps1 for windows users

SkillGG:


package.json
```diff
@@ -10,7 +10,7 @@
         "generate": "ts-node ./scripts/multisrc/generate.ts",
         "clearMulti": "ts-node ./scripts/clearMultisrc.ts",
         "less": "npx less ./test_web/static/css/index.less ./test_web/static/css/index.css",
-        "host": "host.sh"
+        "host": "chmod +x ./host.sh && ./host.sh"
```
nyagami, line 13: image
somehow my windows cant run this

SkillGG: Try changing the operation concatenation symbol to `;` instead of `&&` and check if it'd work

K1ngfish3r: Can this be changed to ``bash ./host.sh``

nyagami: bash is not in windows

SkillGG: Nom is weird especially when doing cross-platform stuff.
It uses cmd.exe on windows and bash on Linux, so IG we'll need to make a .bat file instead and run it in windows. Well still need to run chmod (for Linux) but make the script itself run regardless if chmod finished?

nyagami: image
doenst work event with `;`

SkillGG: Ye, stupid cmd/sh distinction
I don't know why it works on my machine... Maybe I have changed the npm shell script and mine runs bash, but the default is cmd and that doesn't have chmod. So I guess we are stuck with two different scripts

SkillGG: https://github.com/LNReader/lnreader-sources/pull/725/commits/e3e920f200fa8ea8f51ec851210836e1657199e7
Added both `host-linux` and `host-windows`

nyagami (approved):

nyagami (approved):

PR #751, opened 2024-02-05
Minimum the host branch

nyagami:
only 1 commit in host branch, no matter how many changes are pushed in master


index.ts
```diff
@@ -71,8 +71,8 @@ app.post("/popularNovels/", async (req, res) => {
             filters,
         });
         res.json(novels);
-    } catch (err) {
-        res.json({ error: err });
+    } catch (err: any) {
+        res.json({ error: err.message });
```
SkillGG, line 75: I think errors should be typed as `unknown` because some extensions may throw just a string `throw "message"` which is faster and easier to read (at least imho). Then we could check the type it is with `"message" in err`, but I think an easier way would be to just cast it to string because `Error` objects are nicely converted to strings.
```ts
try{
    throw "abc"; // errMessage = "abc"
    throw new Error("abc") // errMessage = "Error: abc"
    throw new SyntaxError("abc") // errMessage = "SyntaxError: abc"
}catch(unknownError:unknown){
    const errorMessage = `${unknownError}`;
   res.json({ error: errorMessage });
}
```a

rajarsheechatterjee (approved):

PR #771, opened 2024-02-13
plugin

Rider21:
fixed:
- all Russian plugins
- freewebnovel
- multiplugins madara, readwn


plugins/russian/renovels.ts
```diff
@@ -77,15 +80,18 @@ class ReN implements Plugin.PluginBase {
...
+    const totalPages =
+      (body.content.branches?.[0]?.count_chapters ||
+        body.content.count_chapters) / 100;
     const branch_id = body.content.branches?.[0]?.id || body.content.id;
     const chapters: Plugin.ChapterItem[] = [];
 
-    for (let i = 0; i < all; i++) {
+    for (let page = 0; page < totalPages; page++) {
+      await delay(1000);
       const chapterResult = await fetchApi(
```
nyagami, line 91: I think you should implement Plugin.PagePlugin for this source.
reference: https://github.com/LNReader/lnreader-sources/blob/beta/plugins/vietnamese/truyenfull.ts

Rider21: 1) the order of chapters will be disrupted
2) some pages will be blank due to filtering

PR #784, opened 2024-02-18
HotNovelPub

Rider21:
Added multi-source HotNovelPub


Rider21:
Replaced `jsmin` with `terser`.
Updated dependencies to the latest versions
added translator separation in ranobelib plugin
added separation of additional chapters in noveltl plugin

terser.json
```diff
@@ -0,0 +1,15 @@
...
+  },
+  "mangle": {},
+  "ecma": "ES5",
+  "enclose": false,
+  "keep_classnames": false,
+  "keep_fnames": false,
+  "ie8": false,
+  "module": false,
+  "safari10": false,
+  "toplevel": false
```
nyagami, line 14:
```suggestion
  "toplevel": true
```

plugins/russian/ranobelib.ts
```diff
@@ -132,10 +151,21 @@ class RLIB implements Plugin.PluginBase {
...
-      novel.chapters = chapters.reverse();
+      novel.chapters =
+        chaptersJson.chapters.branches?.length &&
+        chaptersJson.chapters.branches.length > 1
+          ? chapters.sort((a, b) => {
+              if ((a.page || 0) > (b.page || 0)) return 1;
+              if ((a.page || 0) < (b.page || 0)) return -1;
+              return (a.chapterNumber || 0) - (b.chapterNumber || 0);
+          })
+          : chapters.reverse();
```
nyagami, line 168: sorting by chapter number is enough. Sorting by page makes no sense if page is provided for chapters

Rider21: Without page sorting, the chapters in the appendix will alternate from one page to another as you read. 
But with page sorting, they move from one page to the next.

scripts/multisrc/hotnovelpub/template.ts
```diff
@@ -0,0 +1,276 @@
...
+          .map((item) => "<p>" + item + "</p>")
+          .join("")
+          .replace(/\n/g, "</p><p>")
+          .replace(/\s/g, " ");
+      }
+    }
+    return chapterText.replace(/\.copy right hot novel pub/g, "");
+  }
+
+  async searchNovels(key_search: string): Promise<Plugin.NovelItem[]> {
```
nyagami, line 135:
```suggestion
  async searchNovels(key_search: string, pageNo: number): Promise<Plugin.NovelItem[]> {
```
because app implemented page for searching so you have to handle it

scripts/json_plugins.ts
```diff
@@ -79,7 +80,7 @@ for (let language in languages) {
       pluginSet.add(id);
     }
     const pluginPath = path.join(langPath, plugin);
-    execSync(`jsmin -l 3 --overwrite ${pluginPath}`);
+    execSync(`terser ${pluginPath} --config-file ${terserCfg} -o ${pluginPath}`);
```
nyagami, line 83: I think we should compress file first (move this line to above of the `require` line)

scripts/multisrc/hotnovelpub/template.ts
Rider21: But the source itself doesn't maintain the pages in search.

scripts/json_plugins.ts
Rider21: I'll redo this via the `terser` api later.

nyagami (approved):

PR #790, opened 2024-02-19
(src updates) ENG #2

K1ngfish3r:
- pawread
- royalroad


plugins/english/novelupdates.ts
```diff
@@ -34,38 +34,41 @@ class NovelUpdates implements Plugin.PluginBase {
...
+        page: number,
+        { showLatestNovels, filters }: Plugin.PopularNovelsOptions<typeof this.filters>
+    ): Promise<Plugin.NovelItem[]> { 
+        let link = this.site;
+
+        const filterConditions = [
+            { check: filters.language.value.length > 0, param: 'org', value: filters.language.value },
+            { check: filters.novelType.value.length > 0, param: 'nt', value: filters.novelType.value },
+            { check: (filters.genres.value.include?.length || 0) > 0, param: 'gi', value: filters.genres.value.include },
+            { check: (filters.genres.value.include?.length || 0) > 0, param: 'mgi', value: filters.genre_operator.value },
```
nyagami, line 47: why the `check` uses `filters.geners` but the `value` uses `filters.genre_operator`

plugins/english/scribblehub.ts
```diff
@@ -34,37 +34,40 @@ class ScribbleHubPlugin implements Plugin.PluginBase {
...
-            filters.content_warning.value.exclude?.length)
-            link += '&mct' + filters.content_warning_operator.value;
-
-        link += '&sort=' + filters.sort.value;
-        link += '&order=' + filters.order.value;
+        let link = this.site;
+
+        const filterConditions = [
+            { check: (filters.genres.value.include?.length || 0) > 0, param: 'gi', value: filters.genres.value.include },
+            { check: (filters.genres.value.include?.length || 0) > 0, param: 'mgi', value: filters.genre_operator.value },
```
nyagami, line 41: the same question with novelupdates

plugins/english/novelupdates.ts
K1ngfish3r: after testing NU series finder I noticed that if only the `include` part of the array is true, we append the `genre_operator`. It doesn't apply to the `exclude` array

nyagami (approved):

nyagami:
please change source branch to master and resolve my comment

PR #803, opened 2024-02-25
Added neobook plugin

Rider21:
Added `neobook` plugin
Added `qs` module
added `expandURL` function (optional function designed to expand url, used in webview).
depends on https://github.com/LNReader/lnreader/pull/975


plugins/russian/ranobelib.ts
```diff
@@ -104,6 +105,10 @@ class RLIB implements Plugin.PluginBase {
       }
     });
 
+    this.ui = loadedCheerio("a.header-right-menu__item")
+      .attr("href")
+      ?.replace?.(/[^0-9]/g, "");
```
nyagami, line 110: this property is alive only from when it is assigned to when app is closed. 
is it really important?

Rider21: Without it, you will not be able to open chapters and some books if you are logged in via `WebView'.

nyagami (approved):

PR #822, opened 2024-03-04
Added database for plugins

Rider21:
Added database, cookie manager for plugins
Added copying from webview localstorage, sessionStorage
Depends on https://github.com/LNReader/lnreader/pull/986


Rider21:
plugins
- ranobelib rewritten for the new version of the site
- author.today added authorization support
- bookriver, novelTl, ranobehub optimized (cheerio replaced by match, replace)

multiplugins
- updated readwn domains

icons
- updated icons (now all icons are compressed and their sizes are 96x96 or smaller)
- added script to update icons of all plugins (it will also remove all unused icons)

PR #867, opened 2024-04-01
Add ElloTL as new Plugin

Batorian:
Chapter content still a bit "raw" but can be fixed over time.


nyagami:
novel and chapter path should not contain the site base url
eg: `/series/a-b-c` instead of `https://site.com/series/a-b-c`

PR #897, opened 2024-04-21
Add NovHell plugin

elfing44:
Add NovHell


plugins/french/novhell.ts
```diff
@@ -0,0 +1,209 @@
...
+        novelName = $(elem).find('figcaption a strong').first().text().trim();
+      }
+      const novelCover = $(elem).find('a img').attr('src');
+      const novelUrl = $(elem).find('a').attr('href');
+
+      if (novelUrl && novelName) {
+        novel = {
+          name: novelName,
+          cover: novelCover,
+          path: novelUrl,
```
nyagami, line 54: novel path doesnt include the site domain
eg: `/novel/1` instead of `https://site.com/novel/1`

nyagami:
Could you make the plugin icon squre?
Also please check this novel, some chapters have the same url
image

elfing44:
> Could you make the plugin icon squre? Also please check this novel, some chapters have the same url image

The error comes from the site which has chapters which do not point to the correct link
I will correct the icon

nyagami:
the icon is square, but too large, it should be around 10kb

nyagami (approved):

PR #903, opened 2024-04-23
plugin fix

### LINKED ISSUE #899: RanobeLib doesnt work at all
### Source information

RanobeLib

### Steps to reproduce

Browse -> RanobeLib
or opening any LN, which was liked from RanobeLib source

### Expected behavior

Source should work as intended

### Actual behavior

Doesn't work at all

### LNReader version

1.1.19

### Android version

Android 13

### Other details

I think it is the cause of website ui update, so parser is messed up

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[1.1.19](https://github.com/LNReader/lnreader/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/LNReader/lnreader/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

Rider21:
Plugins
bookriver, novelTl, ranobehub optimized (cheerio replaced by match, replace)
ranobelib temporarily works on old version (closes #899)

Multiplugins
Readwn domains updated, each filter moved to a separate json


nyagami:
Some plugins need version to be updated even it is just refactoring code.
Do hotnovelpub plugins still work with `/` added in the baseUrl. Current app doenst update novel and chapter paths

PR #905, opened 2024-04-24
Add harkeneliwood french + Updated duplicate check for chapter paths.

elfing44:


test_web/static/js/index.js
```diff
@@ -860,7 +860,20 @@ class PluginWrapper {
...
+        });
+
+        const duplicatePaths = Object.keys(pathCountMap).filter(
+          path => pathCountMap[path] > 1,
+        );
+
+        if (duplicatePaths.length > 0) {
+          alert('Duplicate chapter paths found: ' + duplicatePaths.join(', '));
+        }
       }
```
nyagami, line 877: You should use Set to get duplicated chapter path for performance and clean code.
hint: 
 - use 2 sets, one for storing chapter paths, one for storing duplicated paths.
 - for each path, check if set1 has it, then put the path in the corresponding set

plugins/french/harkeneliwood.ts
```diff
@@ -0,0 +1,184 @@
...
+  async getAuthor(text: string): Promise<string> {
+    const regexAuteur = /Auteur\s*:\s*(.*?)\s*(?:\r?\n|$)/i;
+    const match = regexAuteur.exec(text);
+
+    if (match !== null && match[1].trim() !== '') {
+      return match[1].trim();
+    }
+
+    return '';
+  }
```
nyagami, line 139:
```suggestion
  getAuthor(text: string){
    const regexAuteur = /Auteur\s*:\s*(.*?)\s*(?:\r?\n|$)/i;
    const match = regexAuteur.exec(text);

    if (match !== null && match[1].trim() !== '') {
      return match[1].trim();
    }

    return '';
  }
```

plugins/french/harkeneliwood.ts
```diff
@@ -0,0 +1,184 @@
...
+          path: chapterUrl.replace(this.site, ''),
+          releaseTime: releaseDate,
+        });
+      }
+    });
+    novel.chapters = chapters;
+    return novel;
+  }
+
+  async getSummary(text: string): Promise<string> {
```
nyagami, line 90:
```suggestion
  getSummary(text: string) {
```

plugins/french/harkeneliwood.ts
```diff
@@ -0,0 +1,184 @@
...
+    novel.status = NovelStatus.Ongoing;
+    let chapters: Plugin.ChapterItem[] = [];
+    $('#content .entry-content p a').each((i, elem) => {
+      const chapterName = $(elem).text().trim();
+      const chapterUrl = $(elem).attr('href');
+      // Check if the chapter URL exists and contains the site name.
+      if (chapterUrl && chapterUrl.includes(this.site) && chapterName) {
+        const releaseDate = dayjs(
+          chapterUrl?.substring(this.site.length + 1, this.site.length + 11),
+        ).format('DD MMMM YYYY');
```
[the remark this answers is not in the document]
elfing44: I took example from chireads.js which formats dates in the same way. And that doesn't seem to be a problem, does it?
[https://github.com/LNReader/lnreader-plugins/blob/master/plugins/french/chireads.ts#L191](url)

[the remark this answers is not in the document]
nyagami: Oh, i think dayjs handles this format

nyagami (approved):

PR #922, opened 2024-05-01
NU Updates and Fixes

Batorian:
- Chapter names are now displayed correctly
- Added `Mirilu – A Novel Reader Attempts Translating` as supported site
- Added `Zetro Translation` as supported site
- Fixed `Genesis Translations` not getting recognized as `WordPress` Generator
- User now gets info again when chapter content cannot be displayed


plugins/english/novelupdates.ts
```diff
@@ -142,8 +142,19 @@ class NovelUpdates implements Plugin.PluginBase {
...
+      'ss': 'SS',
+    };
+
     loadedCheerio('li.sp_li_chp').each((i, el) => {
-      const chapterName = loadedCheerio(el).text().trim();
+      let chapterName = loadedCheerio(el).text();
+      for (let name in nameReplacements) {
+        chapterName = chapterName.replace(name, nameReplacements[name]);
+      }
+      chapterName = chapterName.replace(/\b\w/g, l => l.toUpperCase()).trim();
```
rajarsheechatterjee, line 157: We could also capitalise `nameReplacements` instead of replacing again, right?

Batorian: Since I also want to capitalise words like `prologue` or `epilogue` without having to add every single possible word to `nameReplacements`, I wanted to first the necessary stuff like `v` to `volume ` or `c` to ` chapter ` and after that has been done, go over the whole chapterName and capitalise everything, not just words I have set as `nameReplacements`.

Batorian (commented):
NU titles sometimes have special characters in it that can't be searched.
-> `*`, `‘`, `’`
These changes replace or remove these special characters in order to enhance the search results.

rajarsheechatterjee (approved):

Issue #860, opened 2024-03-26
Linovelib: Chapters more than 2 pages only show the first page


czqdavid:
### Source information

Linovelib (1.1.1)

### Steps to reproduce

1. Open Linovelib  and open the first novel, 欢迎来到实力至上主义的教室 
2. Open the 4th one in chapter list, 第一卷 — 欢迎来到梦幻般的校园生活
3. Open the webview of this chapter

### Expected behavior

Chapter should be the same with the one in webview.

### Actual behavior

It only shows the first page of the chapter in webview. It seems that it fails to parse the next page.

### LNReader version

v2.0.0-beta.0

### Android version

Android 10 

### Other details

I also try other chapters more than 2 pages in other novels. They all have the problem. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[1.1.12](https://github.com/LNReader/lnreader/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/LNReader/lnreader/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

czqdavid:
I checked the code of plugin and found out next page checking didn't work for me. In the code it checks "下一页" (Chinese Simplified) string but in webview i only found "下一頁"(Chinese Traditional) instead.   My smartphone is in Chinese Simplified and the site is in Chinese Simplified. It's weird that only NextPage is in Chineses Traditional. I think just add  "下一頁" for next page check may solve my issue

czqdavid:
Besides, i also found a section of code for searching  is commented out. Now i also have issue with searching. I only got blank images for result. 
If i clicked the blank image, it showed comments of the novel. There was only one blank image with author's name below. But if i clicked that one, it was keeping loading.
Generally search doesn't work for me.

nyagami:
you have to increase the plugin version (1.1.2)

nyagami:
For search issue:
open source in webview -> search a keyword -> bypass by click on the button -> restart app

PR #925, opened 2024-05-05
Fix Linovelib

czqdavid:
* fix  [#860](https://github.com/LNReader/lnreader-plugins/issues/860)

* fix next page check in Linovelib

* fix incorrect next page url if chapter has more than 3 pages in Linovelib


nyagami:
you have to increase the plugin version (1.1.2)

nyagami (approved):

PR #927, opened 2024-05-06
icon_reloading update

error7404:
icon_reloading update: 49b813fb
 - skip list as string list
 - better root dir, ts-node compatibility
 - parsing plugins.json instead of parsing folders
 - check if google returns notavailable icon
 - better log

added `npm run icons`


scripts/icon_reloading.ts
```diff
@@ -4,78 +4,112 @@ import { Plugin } from '@typings/plugin';
...
 
-        if (!skip[id] && icon && site) {
-          const image = await fetch(
-            `https://www.google.com/s2/favicons?domain=${site}&sz=${size}&type=png`,
-          )
-            .then(res => res.arrayBuffer())
-            .then(res => Buffer.from(res));
+        const notAvailableImage = fs.readFileSync(
+          path.join(root, 'scripts', 'notavailable.png'),
+        );
```
nyagami, line 85: This can be reused by declareing it before the loop

nyagami (approved):

PR #934, opened 2024-05-14
xiaowaz update and fixs

error7404:
Xiaowaz: c78c15a
 - get all novels from all categories
 - cover on search/popular
 - better image finding
 - fix summary

097998e
 - more autor regex patterns
 - restored 'Rebirth of The Thief Who Roamed The World' weird fields
 - `return false;` instead of `return;`  ¯\\\_(ツ)_/¯

84451d7
 - Static novels list, for less fetch
 - Popular novels split based on articles count instead of page max number
 - async images fetching
 - added status to cancelled novels
 - added original tag

### DIFF
--- a/plugins/french/xiaowaz.ts
+++ b/plugins/french/xiaowaz.ts
@@ -1,4 +1,4 @@
-import { CheerioAPI, load } from 'cheerio';
+import { Cheerio, CheerioAPI, load, Element } from 'cheerio';
 import { fetchApi, fetchFile } from '@libs/fetch';
 import { Plugin } from '@typings/plugin';
 import { Filters, FilterTypes } from '@libs/filterInputs';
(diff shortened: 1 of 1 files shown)


plugins/french/xiaowaz.ts
```diff
@@ -85,35 +107,35 @@ class XiaowazPlugin implements Plugin.PluginBase {
...
-      }
+
+      if (PARAGRAPH_EXCLUDE_LIST.some(keyword => textbalise.includes(keyword)))
+        return;
+
       if (
         !textbalise.includes('Genre') &&
         !textbalise.includes('Synopsis') &&
-        //Managing the novel 'Rebirth of The Thief Who Roamed The World' to remove these three fields.
-        !textbalise.includes('重生之賊行天下') &&
```
elfing44, line 109: it was to avoid seeing the title of the novel in the summary

error7404: i forgot i removed it... oops

plugins/french/xiaowaz.ts
```diff
@@ -10,52 +10,74 @@ class XiaowazPlugin implements Plugin.PluginBase {
...
-      (i, elem) => {
-        let novelName = $(elem).first().text().trim();
-        if (novelName.charAt(novelName.length - 1) === '✔') {
-          novelName = novelName.slice(0, -1);
-        }
+    let $ = await this.getCheerio(this.site);
+
+    const categories: Cheerio<Element>[] = [
+      $('li.page_item').find('a:contains("Séries")').parent(),
+      $('li.page_item').find('a:contains("Créations")').parent(),
```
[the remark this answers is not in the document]
error7404: done

elfing44 (approved):

plugins/french/xiaowaz.ts
```diff
@@ -10,52 +10,74 @@ class XiaowazPlugin implements Plugin.PluginBase {
...
+      novel.cover =
+        $novel('img[fetchpriority = "high"]').first().attr('src') ||
+        $novel('img.aligncenter').first().attr('src') ||
+        defaultCover;
+    }
     return novels;
   }
 
+  async popularNovels(pageNo: number, {}: any): Promise<Plugin.NovelItem[]> {
+    const maxPages = 4;
```
nyagami, line 66: I think you should use pageSize instead. 
example:
```js
const pageSize = 10;
const maxPage = Math.ceil(novels.length/pageSize);
if (pageNo  > maxPage) return [];
```

error7404: I wanted to return before fetching all the novels, but I guess I can do the same with a static, I will make even less requests

nyagami (approved):

PR #946, opened 2024-05-21
Update Agitoon.ts for domain update

jake1015:
Plugin is not working because Domain was changed


nyagami:
It should be version 2.0.0

jake1015:
i updated plugin version to 2.0.0. Thank you

nyagami (approved):

PR #955, opened 2024-05-26
Fix hardcoded domain in novel search

jake1015:
Modified the novel search feature to use this.site instead of a hardcoded domain.


plugins/korean/Agitoon.ts
```diff
@@ -9,7 +9,7 @@ class Agitoon implements Plugin.PluginBase {
   name = 'Agitoon';
   icon = 'src/kr/agitoon/icon.png';
   site = 'https://agit655.xyz';
-  version = '2.0.0';
+  version = '3.0.0';
```
nyagami, line 12: this is a minor fix -> 2.0.1

jake1015: okay

PR #967, opened 2024-05-30
feat: Add GitHub action theme auto label

error7404:
Github action to check if a source request has Madara theme or lighnovelwp theme and automatically add the correct label to it (more multisrc can be added but would need to find unique identifiers)

image

---

will add a comment when the action cannot access the URL that was sent

image

### DIFF
--- a/.github/workflows/theme_auto_label.yml
+++ b/.github/workflows/theme_auto_label.yml
@@ -1,0 +1,138 @@
+name: Theme auto label
+
+on:
+  issues:
+    types: [labeled]
+
+
+jobs:
+  auto-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if label is "Source Request"
+        id: label_check
+        run: |
+          LABEL_NAME=$(jq --raw-output .label.name "$GITHUB_EVENT_PATH")
+          if [[ "$LABEL_NAME" == "Source Request" ]]; then
+            echo "source_request=true" >> $GITHUB_OUTPUT
+          else
+            echo "source_request=false" >> $GITHUB_OUTPUT
+            echo "# ❌ Skipping: label is not Source request" >> $GITHUB_STEP_SUMMARY
+          fi
+
+      - name: Get Source Request URL
+        if: steps.label_check.outputs.source_request == 'true'
+        id: get_url
+        run: |
+          ISSUE_BODY=$(jq --raw-output .issue.body "$GITHUB_EVENT_PATH")
+          URL=$(echo "$ISSUE_BODY" | sed -n '/### Source link/,+2p' | sed '1,2d' | tr -d '[:space:]')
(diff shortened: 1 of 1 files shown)


.github/workflows/theme_auto_label.yml
```diff
@@ -0,0 +1,138 @@
...
+      - name: Check if label is "Source Request"
+        id: label_check
+        run: |
+          LABEL_NAME=$(jq --raw-output .label.name "$GITHUB_EVENT_PATH")
+          if [[ "$LABEL_NAME" == "Source Request" ]]; then
+            echo "source_request=true" >> $GITHUB_OUTPUT
+          else
+            echo "source_request=false" >> $GITHUB_OUTPUT
+            echo "# ❌ Skipping: label is not Source request" >> $GITHUB_STEP_SUMMARY
+          fi
```
nyagami, line 21:
```suggestion
    if: contains(github.event.issue.labels.*.name, 'Source Request')
    runs-on: ubuntu-latest
    steps:
```
https://docs.github.com/en/actions/learn-github-actions/expressions#example-using-an-object-filter

.github/workflows/theme_auto_label.yml
```diff
@@ -0,0 +1,138 @@
...
+
+      - name: Get Source Request URL
+        if: steps.label_check.outputs.source_request == 'true'
```
nyagami, line 24:
```suggestion
```

.github/workflows/theme_auto_label.yml
```diff
@@ -0,0 +1,138 @@
...
+        id: get_url
+        run: |
+          ISSUE_BODY=$(jq --raw-output .issue.body "$GITHUB_EVENT_PATH")
```
nyagami, line 27: you can access issue body via context `github.event.issue.body`

.github/workflows/theme_auto_label.yml
error7404: I've read it but my brain never caught it...

.github/workflows/theme_auto_label.yml
```diff
@@ -0,0 +1,138 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          GITHUB_ISSUE_NUMBER: ${{ steps.get_issue_number.outputs.GITHUB_ISSUE_NUMBER }}
+        run: |
+          if [[ -z "$URL" ]]; then
+            echo "url_valid=false" >> $GITHUB_OUTPUT
+            echo "# ❌ Error: URL not found" >> $GITHUB_STEP_SUMMARY
+            exit 1
+          else
+            if curl --output /dev/null --silent --head --fail "$URL"; then
```
nyagami, line 53: I think adding an user agent in headers could increase accessibility

nyagami (approved):

PR #986, opened 2024-06-11
Agitoon domain update

jake1015:
domain number 657 -> 658


nyagami:
Btw, you can fix current error by open agiton in webview -> bypass cloudflare -> close webview -> restart app

nyagami (approved):

PR #1003, opened 2024-06-25
Add TopLiba

Rider21:
added source `TopLiba`
deleted unused `fetchFile` import
replaced fetch with `fetchApi`
added `eslint`
fixed `multiplugin` generation


package.json
```diff
@@ -16,7 +16,7 @@
         "host-windows": "powershell ./host.ps1",
         "prettier": "prettier --write \"./**/*.{js,ts}\"",
         "prepare": "husky install",
-        "icons": "npm run clearMultisrc && npm run generate && npx tsc -m node16 && npm run json && node ./scripts/icon_reloading.js"
```
error7404, line 19: icon_reloading needs the json file to be generated, and it would be better if the multisrc was generated before it

nyagami (approved):

PR #1030, opened 2024-07-06
jest tests for plugins

CD-Z:
Helps get a general overview which plugins work and which do not.

However it is not completely consistent. For example:

- LightNovelPub:
  - works in app
  - doesn't work in web
  - works partly in tests
- RanobeLib:
  - doesn't install in app
  - works in web
  - doesn't work in tests 

I tried to figure out why this is, but wasn't able to.

### DIFF
--- a/scripts/multisrc/readwn/sources.json
+++ b/scripts/multisrc/readwn/sources.json
@@ -19,18 +19,18 @@
   },
   {
     "id": "wuxiamtl",
-    "sourceSite": "https://www.fansmtl.org",
+    "sourceSite": "https://www.fanmtl.com",
     "sourceName": "Fans MTL",
     "options": {
-      "versionIncrements": 1
+      "versionIncrements": 2
     }
   },
   {
     "id": "fannovel",
-    "sourceSite": "https://www.fannovels.org",
+    "sourceSite": "https://www.fanmtl.com",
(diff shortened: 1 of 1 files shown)


nyagami:
When we should run this script? Is it better than just waiting for user's report when some plugins not working

CD-Z:
We could exclude them, then we would at least had a few automted or just close the pr. The main reason i created it was because i wanted to work a bit with jest.

nyagami:
Imo unit tests is used for functionalities testing.

It should have test cases where plugin passed (success or fail as expected).

And the test input data should be staticly prepared. Which means you just need to provide some random html files or json object via a local server (could be a proxy).

I think you are testing both plugin functionalities and availability.

nyagami:
Imo unit tests is used for functionalities testing.

It should have test cases where plugin passed (success or fail as expected).

And the test input data should be staticly prepared. Which means you just need to provide some random html files or json object via a local server (could be a proxy).

I think you are testing both plugin functionalities and availability.

nyagami:
But if you still want to test plugin availability, i recommend using some tools like selenium or puppeteer.

CD-Z:
We could exclude them, then we would at least had a few automted or just close the pr. The main reason i created it was because i wanted to work a bit with jest.

PR #1060, opened 2024-07-15
feat: [Novel Updates]: Automatic List Tracking

Batorian:
- when you click a chapter, it gets automatically marked in Novel Updates as read
- since the path changes, all Novel Updates chapter entries need to be cleared and parsed again
- the path now includes the novelId and chapterId
- the path gets split into 3 parts later on: novelId, chapterId, chapterPath


rajarsheechatterjee:
If a user downloads all the chapters, they would also be marked as read, which wouldn't be accurate, right?

Batorian:
> If a user downloads all the chapters, they would also be marked as read, which wouldn't be accurate, right?

Yeah, that wouldn't be ideal. Would probably need to add more parameters to parseChapter.

PR #1065, opened 2024-07-16
feat: Add Webnovel

Batorian:


src/plugins/english/webnovel.ts
```diff
@@ -0,0 +1,313 @@
+/* eslint-disable no-case-declarations */
```
nyagami, line 1: what is this for

Batorian: Copied over from my template. I removed it.

nyagami (approved):

PR #1077, opened 2024-07-21
Move to Zustand and add Filters to web_test

SkillGG:
- Moved from Redux to Zustand
- Add Filters to Popular novels accordion
- Add rule for consistent type definitions
- Add TriState Switches and Checkboxes (used in XCheckboxFilter)


src/prettify.tsx
```diff
@@ -0,0 +1,7 @@
+export type Prettify<T> = T extends object
+  ? {
+      [K in keyof T]: T[K];
+    } & {
+      /** */
+    }
+  : T;
```
nyagami, line 7: what does this file do?

src/store/pluginStore.ts
```diff
@@ -0,0 +1,27 @@
+import { Plugin } from '@typings/plugin';
+import { StoreCreator } from '.';
+import { getPlugin } from '@provider/plugins';
+
+export type PluginStore = {
+  selectedPlugin?: Plugin.PluginItem;
```
nyagami, line 6:
```suggestion
  pluginItem?: Plugin.PluginItem;
```

src/libs/filterInputs.ts
```diff
@@ -11,59 +17,84 @@ export enum FilterTypes {
...
+type ExcludableCheckboxFilter = {
   type: FilterTypes.ExcludableCheckboxGroup;
   options: readonly FilterOption[];
-  value: { include?: string[]; exclude?: string[] };
-}
+  /** Default value */
+  value: {
+    /** Checkboxes marked as included */
+    include?: string[];
+    /** Checkboxes marked as included */
```
nyagami, line 52:
```suggestion
    /** Checkboxes marked as excluded */
```

src/libs/filterInputs.ts
```diff
@@ -11,59 +17,84 @@ export enum FilterTypes {
...
+  ? CheckboxFilter
+  : Type extends FilterTypes.Picker
+    ? PickerFilter
+    : Type extends FilterTypes.Switch
+      ? SwitchFilter
+      : Type extends FilterTypes.TextInput
+        ? TextFilter
+        : Type extends FilterTypes.ExcludableCheckboxGroup
+          ? ExcludableCheckboxFilter
+          : never);
```
nyagami, line 78: image

Can we make a map type?
Like this
```ts
type FilterMapping = {
  [FilterTypes.CheckboxGroup]: CheckboxFilter
}
export type Filter<Type extends FilterTypes> = {
  label: string;
} & FilterMapping[Type]
```

src/prettify.tsx
SkillGG: https://www.totaltypescript.com/concepts/the-prettify-helper

I thought it'd work on those big types, but it doesn't, unfortunately. I forgot to delete it.

src/libs/filterInputs.ts
SkillGG: I think it should work, but I'd need to test it.

src/store/overlayStore.ts
```diff
@@ -0,0 +1,30 @@
...
+      ...state,
+      loading: true,
+    }));
+  },
+  hideLoading() {
+    set(state => ({
+      ...state,
+      loading: false,
+    }));
+  },
```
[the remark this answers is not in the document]
SkillGG: zustand is global store tho. `useAppStore(state=>state.loading)`.

did you mean local component state?

src/store/pluginStore.ts
SkillGG: You have to change it in other files that use this car as well.

src/store/overlayStore.ts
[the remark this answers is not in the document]
nyagami: I meant global loading. But seems it is not very important. We can remove it xD

src/libs/filterInputs.ts
nyagami: ye, just more readable code, nested code always scares people

PR #1106, opened 2024-08-02
Freewebnovel: Regex to remove website spam

Palloxin:
It removes just a part of the spam.
EDIT: I tested that script with 120+ chapters on chapters published 1+ year ago. I didnt find any more spam.
I also checked recent chapters; The script still deletes that spam effectively.

### DIFF
--- a/src/plugins/english/freewebnovel.ts
+++ b/src/plugins/english/freewebnovel.ts
@@ -116,7 +116,8 @@ class FreeWebNovel implements Plugin.PluginBase {
 
     const chapterText = loadedCheerio('div.txt').html() || '';
     return chapterText
-      .replace(/[ƒf][rɾг][ēeё][ēёe][wω][ёēe][bɓ][nɳη][ø૦ѳσo][ѵv][ёeē][lℓɭ]\.\s?[ƈcç][૦σøoѳ][๓ɱm]\.?/gi, '');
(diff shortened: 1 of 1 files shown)


src/plugins/english/freewebnovel.ts
```diff
@@ -115,7 +115,8 @@ class FreeWebNovel implements Plugin.PluginBase {
       loadedCheerio('div.txt').find('p:last-child').remove();
 
     const chapterText = loadedCheerio('div.txt').html() || '';
-    return chapterText;
+    return chapterText
+      .replace(/[ƒf][rɾг][ēeё][ēёe][wω][ёēe][bɓ][nɳη][ø૦ѳσo][ѵv][ёeē][lℓɭ]\.\s?[ƈcç][૦σøoѳ][๓ɱm]\.?/gi, '');
```
nyagami, line 119: Does this kind of string have a specific tag or classname?

Palloxin: I am not sure. The website first loads the spam then removes it. Maybe to tackle web scrapers.

K1ngfish3r: > Does this kind of string have a specific tag or classname?

There is no tag/class name, it is baked directly into the <p> tag

nyagami (approved):

Palloxin:
I think it's ready now

PR #1126, opened 2024-08-09
RoyalRoad:  Change parseChapter to use regex

K1ngfish3r:
Requires `style` tag to be bypassed in the sanitizer for all text formatting to work properly
But this fixes all/most missing text


nyagami:
undefined at the start of content.
path: fiction/90655/worm-mage/chapter/1729145/chapter-2-wormhole-core
image

K1ngfish3r:
updated

nyagami (approved):

Issue #78, opened 2021-10-13
Rewayat Club


esraahassan2151:
### Source name

Rewayat Club

### Source link

https://rewayat.club/

### Language

Arabic

deku99hh:
Why is the source not uploaded?

Khalilr28:
There is a source in the Arabic language that is not present in the program, and it is very well known. I hope you add it, as it is the only thing missing in the program for me. The name of the site is.
rewayat.club.

PR #1127, opened 2024-08-10
New Source: Add New Arabic Source: rewayat.club

Dredd1987:
Added a new source plugin for rewayat.club, providing access to Arabic novels. This PR resolves issue #78


nyagami:
I think you should create a new branch from lnreader-plugins/master to avoid duplicated changes.

You can add it easily with Github web GUI 
image

PR #1135, opened 2024-08-12
Improvement: [Novel Updates]: Replace API Fetching with Table Fetching

Batorian:
- fixes chapter url issue


nyagami:
Could you test updating library with 10 NU novels which has at least 400 chapters.

Batorian:
> Could you test updating library with 10 NU novels which has at least 400 chapters.

I'll try when I have time but I think it causes problems. Maybe I will limit simultaneous fetching to 20 pages max. So when a novel has 35 pages or something we first fetch 20 pages at the same time and after that the last 15 pages. Maybe that works better

PR #1145, opened 2024-08-17
New Arabic source sunovels.com

Dredd1987:
Adding the new source, sunovels.com, has resolved issue #130.

However, there's a problem with images on the library and search pages. It takes a while for the requested URLs to show up in the HTML. There might be a more efficient way to get these images, maybe through an API or some other method. I haven’t found a good solution.


K1ngfish3r:
image
get this script tag and do the json magic again?

rajarsheechatterjee:
@Dredd1987 Thank you for contributing to the Arabic plugins. ✨

Dredd1987:
I couldn't figure out the HTTP endpoint for the JSON, and the JSON within the script tag is a hassle to clean up, so I used regex and matching. It's not elegant, but it works.

src/plugins/arabic/sunovels.ts
```diff
@@ -0,0 +1,306 @@
+import { CheerioAPI, load as parseHTML } from 'cheerio';
+import { fetchApi } from '@libs/fetch';
+import { Plugin } from '@typings/plugin';
+import { Filters, FilterTypes } from '@libs/filterInputs';
+import { defaultCover } from '@libs/defaultCover';
+
+class RewayatClub implements Plugin.PagePlugin {
```
K1ngfish3r, line 7:
```suggestion
class Sunovels implements Plugin.PagePlugin {
```
Wrong plugin :)

src/plugins/arabic/sunovels.ts
```diff
@@ -0,0 +1,306 @@
...
+        { label: 'مكتمل', value: 'Completed' },
+        { label: 'جديد', value: 'New' },
+        { label: 'مستمر', value: 'Ongoing' },
+      ],
+      type: FilterTypes.Picker,
+    },
+  } satisfies Filters;
+}
+
+export default new RewayatClub();
```
K1ngfish3r, line 299:
```suggestion
export default new Sunovels();
```
yes

rajarsheechatterjee (approved):

PR #1147, opened 2024-08-18
add rainofsnow english

sweetbbak:
I added rainofsnow. I only realized til afterwords that apparently NovelUpdates has some parsing logic for this site, but I still think it would be nice to have as it is one of my favorite sites.

One issue I was having was when I would test the scraper using `npm start` and when I would parse a Novel's chapters, I would have a link that looks like this:
`https://rainofsnow.com/chapters/missing-fiance-ch1/?novelid=61402`

npm says I cant have the full link here, so i slice it off with `?.slice(novelPath.length);` and get 'chapters/missing-fiance-ch1/?novelid=61402' which testing this with 'Parse Chapter' crashes the proxy with 'Invalid URL' but if I use the full link, it works perfectly fine... but then the test site complains that it needs to be relative. Though, *no* combination of the url segments work except for the full URL. Im a Golang dev so this is a little foreign to me to begin with, but this was a little baffling to me.

side note: I also meant to ask how hosting my own repository would work? I try to add a repository URL in 'settings > repositories > add repo' but it returns invalid no matter what. What exactly is it asking for here? what does the git repo structure supposed to be like? I can write some docs on this if I can figure it out. Thanks for the awesome application.


K1ngfish3r:
LNReader tacks on `this.site` for every link, so you can do `.slice(this.site.length)` when inserting into the db

To host your own repo, run `npm run host-windows` or `npm run host-linux` then take the raw link of dist/plugins.min.json inside the plugins branch that gets created on your repo

src/plugins/english/rainofsnow.ts
```diff
@@ -0,0 +1,204 @@
...
+        // ok sooo wtf. npm says I cant have the full link here, so i slice it off with
+        // ?.slice(novelPath.length);
+        // and get 'chapters/missing-fiance-ch1/?novelid=61402'
+        // which testing this with 'Parse Chapter' crashes the proxy with 'Invalid URL'
+        // but if I use the full link, it works perfectly fine... but then the test site complains that it needs to be relative -_-
+
+        if (path && name) {
+          novelChapters.push({ name, path, releaseTime });
+        }
+      });
```
K1ngfish3r, line 110:
```suggestion
    loadedCheerio('#chapter')
      .find('li')
      .each((i, el) {
        const name = loadedCheerio(el).find('.chapter').first().text().trim();
        const releaseTime = loadedCheerio(el).find('small').text();
        const path = loadedCheerio(el).find('a').attr('href').slice(this.site.length);

        // the link looks like this:
        // https://rainofsnow.com/chapters/missing-fiance-ch1/?novelid=61402
        //
        // ok sooo wtf. npm says I cant have the full link here, so i slice it off with
        // ?.slice(novelPath.length);
        // and get 'chapters/missing-fiance-ch1/?novelid=61402'
        // which testing this with 'Parse Chapter' crashes the proxy with 'Invalid URL'
        // but if I use the full link, it works perfectly fine... but then the test site complains that it needs to be relative -_-

        if (path && name) {
          novelChapters.push({ name, path, releaseTime });
        }
      });
```
You did it once before for `parseNovels`

src/plugins/english/rainofsnow.ts
```diff
@@ -0,0 +1,204 @@
+import { fetchApi } from '@libs/fetch';
+import { Plugin } from '@typings/plugin';
+import { Filters } from '@libs/filterInputs';
+import { CheerioAPI, load as parseHTML } from 'cheerio';
+import { NovelStatus } from '@libs/novelStatus';
+
+const baseUrl = 'https://rainofsnow.com/';
+
+class Rainofsnow implements Plugin.PluginBase {
```
K1ngfish3r, line 9:
```suggestion
class Rainofsnow implements Plugin.PagePlugin {
```
Needs to be a PagePlugin, I'm afraid

sweetbbak:
Thats what I had figured. The thing that was throwing me for a loop was that the test site would work for parsing the popular page and for "Search" using `.slice(this.site.length)` but using that URL slug would cause an error when trying to parse the chapters contents. Im not sure exactly why, but it was only working with the full URL. It would crash the test proxy server with `ERR_INVALID_URL` so I was at a loss.

I end up getting this:
```
[1]   code: 'ERR_INVALID_URL',
[1]   input: 'chapters/reaper-c3/?novelid=61654'
[1] npm run proxy exited with code 1
```

nyagami:
Hi thank for your contributing. 
To be clear: App doesnt store full url, because if the domain get changed users data would be lost.
So you just need to do `this.site + chapter Path` in parse chapter to get full url.

Also you could use earlynovel (en) as reference for implementing a site have paged chapters

sweetbbak:
Thanks! I'll work on that, make sure its working correctly, and amend the PR sometime tomorrow.

sweetbbak:
Alright, everything is working well. The plugin type is now correct, it can parse pagination without an issue, along with everything else. I tested the resulting chapter HTML in my browser and it looks good as well (though I am not familiar with how that is rendered in the app). I also cleaned things up a bit.

src/plugins/english/rainofsnow.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  async searchNovels(
+    searchTerm: string,
+    pageNo: number,
+  ): Promise<Plugin.NovelItem[]> {
+    const novels: Plugin.NovelItem[] = [];
+    // get novels using the search term
+    // no page number, infinite scroll
+
```
nyagami, line 156: You should return undefined or empty array if page > 1. That is a signal for end of infinite fetching.

PR #1159, opened 2024-08-21
Faqwiki: Refactor

K1ngfish3r:
Page changed a lot, will probably break previously available links (if any)
Closes https://github.com/LNReader/lnreader-plugins/issues/1152


src/plugins/english/faqwikius.ts
```diff
@@ -28,7 +31,8 @@ class FaqWikiUs implements Plugin.PluginBase {
         .find('a')
         .attr('href')
         ?.replace('tp:', 'tps:')
-        ?.slice(this.site.length);
+        ?.slice(this.site.length)
+        ?.slice(0, -1);
```
nyagami, line 35: unsafe tho, you should use replace

nyagami (approved):

PR #1162, opened 2024-08-22
Update Novelbin

CD-Z:
adds covers to search results
removes novelbin source tags from chaptertext


src/plugins/english/novelbin.ts
```diff
@@ -7,8 +7,8 @@ class NovelBin implements Plugin.PluginBase {
   id = 'novelbin';
   name = 'Novel Bin';
   icon = 'src/en/novelbin/icon.png';
-  site = 'https://binnovel.com/';
-  version = '1.0.1';
+  site = 'https://lightnovel.novelupdates.net/';
```
nyagami, line 10: This site redirects me to https://novelbin.com/

nyagami: So I think you should use `https://lightnovel.novelupdates.net/` for building chapter url instead of changing the domain

PR #1170, opened 2024-08-27
new source dilar.tube

Dredd1987:
I got stuck on search and chapter parsing. Maybe you can help.


hoangquan0511:
Where did you get stuck?. Like cloudflare issue or understanding language issue

Dredd1987:
I fail to fetch the response of search, the entire novels novelsearch function doesn't work, as for the chapterText I think it pulls the page before the text being populated. And the search response is encrypted I think a got It right but don't know if can be done in the plugin it self or it gonna severely affect the preformance

Dredd1987:
It's a bit difficult to sort through the site am translating as I go

hoangquan0511:
I see that they use quick_search api entry for searching in app bar. Have you tired it?

K1ngfish3r:
image
```ts
const formData = new FormData();
formData.append('query', 'wife');
formData.append('includes', '["Manga","Team","Member"]');
fetch('https://dilar.tube/api/quick_search', {
    method: 'POST',
    body: formData,
}).then(a => a.json()).then(b => console.log(b));
```
edit as needed

K1ngfish3r:
What's current status?

Rider21:
By default, the `crypto` module is missing from `react-native`.
Have you found a replacement for it?
For example, `expo-crypto`, `react-native-quick-crypto`, `crypto-browserify`.

Dredd1987:
> What's current status?

Sorry, I got really busy. The chapter content is still returning an error with the connection being refused.

Dredd1987:
> By default, the `crypto` module is missing from `react-native`. Have you found a replacement for it? For example, ~`expo-crypto`~, `react-native-quick-crypto`, `crypto-browserify`.

I abandoned this approach because of a CORS error. It works in my tests, but not at all in the app.

rajarsheechatterjee (approved):

PR #1171, opened 2024-08-27
galaxytranslations new url

r1di:
galaxytranslations new url


nyagami:
you need to change versionIncrement

nyagami (approved):

PR #1208, opened 2024-10-02
fix novelupdates

nguyd1:


Batorian:
Have you considered that NU is adding chapters to the pages reversed? So every time a chapter is added to a novel, you'll need to update all pages. That's the main reason why we still use admin-ajax + the long fetch times.

nguyd1:
> Have you considered that NU is adding chapters to the pages reversed? So every time a chapter is added to a novel, you'll need to update all pages. That's the main reason why we still use admin-ajax + the long fetch times.

hey, not sure what u mean exactly by pages, are u referring to the pages on the novelupdates website? i was testing with fresh novels (deleted cache so first time fetching all chapters) and i was still getting the wrong chapters, looked at the data i was getting from ajax and i was receiving duplicate chapter urls pointing to chapters from different novels

src/plugins/english/novelupdates.ts
```diff
@@ -6,7 +6,7 @@ import { Plugin } from '@typings/plugin';
 class NovelUpdates implements Plugin.PluginBase {
   id = 'novelupdates';
   name = 'Novel Updates';
-  version = '0.7.9';
+  version = '0.8.7';
```
rajarsheechatterjee, line 9:
```suggestion
  version = '0.8.0';
```

lalalasupa0:
https://github.com/LNReader/lnreader-plugins/issues/1480 Related it seems, I tested your plugin right quick but nuh huh

Screenshot_2025-02-13_090741_com rajarsheechatterjee LNReader
Screenshot_2025-02-13_090248_com rajarsheechatterjee LNReader
Screenshot_2025-02-13_090344_com rajarsheechatterjee LNReader
Screenshot_2025-02-13_090400_com rajarsheechatterjee LNReader

lalalasupa0:
Actually, the NU website itself has had issues with bad linking within the TOCs recently, so maybe this is all related to it ?
For reference: https://www.novelupdatesforum.com/posts/7885250/

Batorian:
I propose closing this PR as implementing pages to NU is not a good solution. It was an approach I considered at first but considering a single page only holds 15 chapters it would lead to massive issues during library updates with big novels, as you would send a request to the NU server every time a new page gets loaded. I am still trying for solutions to avoid the chapter duplication issue but introducing paging is sadly not a solution.

rajarsheechatterjee:
Closing this PR as it appears outdated.

If the changes are still relevant, feel free to reopen the PR or submit a new one.

*Note: This was done as part of an automated cleanup of stale pull requests.*

PR #1255, opened 2024-10-29
Update for arnovel url

Abdalrhman1235:


nyagami:
versionIncrements should be positive integer

nyagami (approved):

Abdalrhman1235:
> versionIncrements should be positive integer

After it being updated it tries to enter the url through bing so it doesn't work what should I do

nyagami:
I forgot to merge it, you can try again

PR #1294, opened 2024-11-04
Dependency optimization

Rider21:
~Removed `urlencode` - not used~
Migration from `qs` to `react-native-url-polyfill` (https://github.com/LNReader/lnreader/pull/1302)


src/plugins/chinese/69xinshu.ts
```diff
@@ -150,7 +150,7 @@ class XinShu69 implements Plugin.PluginBase {
 
     const searchUrl = `${this.site}/modules/article/search.php`;
     const formData = new FormData();
-    // formData.append('searchkey', encode(searchTerm, 'gbk'));
+    formData.append('searchkey', encode(searchTerm, 'gbk'));
```
nyagami, line 153: Does the search work properly in app?

Rider21: I haven't had a chance to check it out yet

Rider21:
I think it's not necessary, react-native already has a `URLSearchParams` function (so replacing `qs` doesn't make any difference)
The update will mostly just fix the loading of illustrations.

src/plugins/chinese/69xinshu.ts
nyagami: Its domain was changed

nyagami (approved):

PR #1320, opened 2024-11-14
Divine Dao Library: update all data fetching

Zegnat:
This introduces filters and the option to check the last updated novels. It also fixes novel covers, which have not been working with the old plugin.

Most of the rewrite was done to correctly report on chapters.

The problem is that DDL adds a bunch of links to chapter pages to their novel page before the chapters go live. E.g. right now Alchemy Master has links up to Chapter 200 but all links over 184 just give a Page not found. The old plugin did not understand this and would show all 200 chapters and was unable to let the reader know when they actually came online. This new plugin will check the latest posts published to the novel-specific website category.

**Screenshots**

image

image

image

### DIFF
--- a/src/plugins/english/divinedaolibrary.ts
+++ b/src/plugins/english/divinedaolibrary.ts
@@ -1,125 +1,259 @@
-import { CheerioAPI, load as parseHTML } from 'cheerio';
-import { fetchApi } from '@libs/fetch';
-import { Plugin } from '@typings/plugin';
 import { defaultCover } from '@libs/defaultCover';
+import { fetchApi } from '@libs/fetch';
+import { Filters, FilterTypes } from '@libs/filterInputs';
+import { Plugin } from '@typings/plugin';
+import { load as parseHTML } from 'cheerio';
 
 class DDLPlugin implements Plugin.PluginBase {
   id = 'DDL.com';
   name = 'Divine Dao Library';
   site = 'https://www.divinedaolibrary.com/';
-  version = '1.0.1';
+  version = '1.1.0';
   icon = 'src/en/divinedaolibrary/icon.png';
 
-  parseNovels(loadedCheerio: CheerioAPI, searchTerm?: string) {
-    let novels: Plugin.NovelItem[] = [];
+  filters = {
+    category: {
+      type: FilterTypes.CheckboxGroup,
+      label: 'State',
+      value: ['Completed', 'Translating', 'Lost in Voting Poll', 'Dropped'],
+      options: [
+        { label: 'Completed', value: 'Completed' },
+        { label: 'Translating', value: 'Translating' },
+        { label: 'Lost in Voting Poll', value: 'Lost in Voting Poll' },
+        { label: 'Dropped', value: 'Dropped' },
+        { label: 'Personally Written', value: 'Personally Written' },
+      ],
+    },
+  } satisfies Filters;
 
-    loadedCheerio('#main')
-      .find('li')
-      .each((i, el) => {
-        const novelName = loadedCheerio(el).find('a').text();
-        const novelCover = defaultCover;
-        const novelUrl = loadedCheerio(el).find('a').attr('href');
-
-        if (!novelUrl) return;
-
-        const novel = {
-          name: novelName,
-          cover: novelCover,
-          path: novelUrl.replace(this.site, ''),
-        };
-
-        novels.push(novel);
-      });
-
-    if (searchTerm) {
-      novels = novels.filter(novel =>
-        novel.name.toLowerCase().includes(searchTerm.toLowerCase()),
-      );
+  /**
+   * Safely extract the pathname from any URL on {@link site}. Check the root
+   * site as there are novels linking off-site (to Patreon).
+   *
+   * @private
+   */
+  getPath(url: string): string | undefined {
+    if (!url.startsWith(this.site)) {
+      return undefined;
     }
+    const trimmed = url.substring(this.site.length).replace(/(^\/+|\/+$)/g, '');
+    if (trimmed.length === 0) {
+      return undefined;
+    }
+    return trimmed;
+  }
+
+  /**
+   * Map an array with an asynchronous function and only return the array items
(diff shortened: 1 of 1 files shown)


src/plugins/english/divinedaolibrary.ts
```diff
@@ -1,125 +1,259 @@
...
+      const linkedChapters = content('li > span > a')
+        .map((_, anchorEl) => {
+          const path = this.getPath(anchorEl.attribs['href']);
+          if (!path) return;
+          return {
+            name: content(anchorEl).text(),
+            path,
+          } satisfies Plugin.ChapterItem;
+        })
+        .toArray();
```
nyagami, line 121: I think you juse need to filter out elements which doesnt have href for chapter.
Slicing chapter list with latest index might have trouble if the site have random releases (eg: .., 99, 100, 125)

Zegnat: This already happens. When you return `undefined` (as is done on line 115) within a Cheerio `.map()` it will [not be part of the resulting array](https://github.com/cheeriojs/cheerio/blob/439fd196cc8e9dc4a4f02113e3275ee0c8fb95cb/src/api/traversing.ts#L684-L686). A Cheerio `.map()` is not like a standard JavaScript `.map()`.

nyagami: So all elements in linkedChapters should have a valid url and slicing this array might be unnecesssary?

tw-martijn: Oh, do you mean the slicing that happens a few lines below this comment? That is a completely separare thing.

The Divine Dao Library website lies about their chapters, as I wrote in the PR body. So what my code does is it parses all the chapter links, which might give links for chapters 1 through 200. Then as a second step in findLatestChapter we can figure out that actually chapter 186 is the last published one (as it is at the time of this comment for [Alchemy Master](https://www.divinedaolibrary.com/alchemy-master/), where any link after chapter 186 is on the page but shows a Not found).

So then the slice simply removes all the links after the one to 186.

Otherwise it would have reported 200 chapters for Alchemy Master and the application has no way to inform the reader about when chapters actually go live. Links would not change at all (even the titles are already in it, e.g. https://www.divinedaolibrary.com/alchemy-master-chapter-192-another-alchemy-duel/ which is a Not found at the time of writing).

Zegnat: I think we are talking about different things. Are you talking about the slice that happens on line 124, just below the lines you left your comment on?

DDL lies about their chapters on novel pages. If you just parse the novel page for [Alchemy Master](https://www.divinedaolibrary.com/alchemy-master/) it will give you 200 **valid** links to 200 chapters. But the latest chapter actually published at time of writing is 186. All links 187 through 200 look like they are there, and even include titles in their URLs, but all result in a Not Found error when you visit them.

I am not slicing faullty links or anything like that. In `this.findLatestChapter` I try to detect the URL for the actually latest published chapter. And then `.slice()` to remove all the links after it on the list, because we are sure those will all just give errors.

This way LNReader can report a correct number of chapters and actually inform readers when new chapters come out. The old plugin was just permanently stuck at showing 200 chapters for Alchemy Master.

**In short:** all URLs are valid, but several URLs at the end of the list are placeholders and need to be cut off.

nyagami: got it

src/plugins/english/divinedaolibrary.ts
```diff
@@ -1,125 +1,259 @@
...
+      );
+    }
+    const novels = (await this.allNovels()).filter(novel => {
+      const selectedCategories = options.filters.category.value;
+      if (Array.isArray(selectedCategories)) {
+        return selectedCategories.includes(novel.category);
+      }
+      return false;
+    });
+    return await this.asyncMap(novels, this.grabNovel.bind(this));
```
nyagami, line 226: I see that you fetch detail for every novel in popular / search to get covers, which is quite expensive. 
I think you could create a temp cache by adding a `Set` in plugin instance for caching novel detail.

nyagami: this would save lot of time if user visit DDL multiple times

Zegnat: Happy to add some caching. Is there a best practice that plugins usually follow? Would you cache the full parsed page (including author and summary) or only the cover URL for display on the novel list?

I felt the few extra requests were worth it to me to make it look better, especially since there are relatively few novels and I thought it unlikely anyone was going to visit it often.

nyagami: You can just store it in a variable (memory) which would be lost when user turn off the app. 

Or you can use plugin storage ([example](https://github.com/LNReader/lnreader-plugins/blob/master/src/plugins/russian/authortoday.ts#L7)). But dont abuse it, storing large amout of data could hurt app performance.

Btw app just need novel name, path, and cover to display novels  in  browse and search screen. So I think you just need to cache cover urls

tw-martijn: Is storage kept between app starts? That does sound like overkill for what is really just an optimisation for the number of requests. I would expect people to rarely go back into the novel list. Optimising for people who are looking to select a novel makes a lot more sense.

Not sure I really like just randomly stuffing information in memory that does not get cleared until application restart either. But I guess the application is keeping an instance of every installed plugin in memory anyway? So it really wouldn’t be that hard.

I’ll have a go at adding a cache for reading the novel pages later today and get this PR updated.

Zegnat: Is storage kept between app starts? That does sound like overkill for what is really just an optimisation for the number of requests. I would expect people to rarely go back into the novel list. Optimising for people who are looking to select a novel makes a lot more sense.

Not sure I really like just randomly stuffing information in memory that does not get cleared until application restart either. But I guess the application is keeping an instance of every installed plugin in memory anyway? So it really wouldn’t be that hard.

I’ll have a go at adding a cache for reading the novel pages later today and get this PR updated.

PR #1341, opened 2024-11-27
Add Komga plugin

NilSilva:
This is a komga plugin. It's not perfect but I think it's good enough for a first version. I'll keep trying to improve it when I can. The biggest problem is the webview in the novel page. Because I'm using the api instead of opening that series page in komga it opens a page with the json api response.

It is using the plugin settings so the user can configure the url, email, and password.


src/plugins/multi/komga.ts
```diff
@@ -0,0 +1,272 @@
...
+        throw new Error(
+          "'btoa' failed: The string to be encoded contains characters outside of the Latin1 range.",
+        );
+      }
+
+      block = (block << 8) | charCode;
+    }
+
+    return output;
+  }
```
Soopyboo32, line 52: issnt btoa and atob global functions already?

src/plugins/multi/komga.ts
```diff
@@ -0,0 +1,272 @@
...
+      const { children, ...rest } = obj;
+      acc.push(rest);
+
+      if (children) {
+        acc.push(...this.flattenArray(children));
+      }
+
+      return acc;
+    }, []);
+  }
```
Soopyboo32, line 65: flat is also already a function on array, id assume hermes supports (couldent find any github posts abt it)?

nyagami: I think hermes does have it. They have tests for this method https://github.com/facebook/hermes/blob/main/test/hermes/array-flat-recursion.js

nyagami: No, it's facebook fauth, they stopped update hermes changelogs. Now only facebook devs can know what JS features is are supported in what react native version

nyagami: wait for new app update, we are going to migrate to new arch of React Native

nyagami: No, but could you write a guide to use this one, I have no idea to use it

nyagami: ye

nyagami:
Follow your guide but i can access my local komga.

Url: http://<lanip>:25600

What did i do wrong?

nyagami:
ye, I have to append "/" and restart app to work.
Is it Komga bug or LNReader bug?

nyagami (approved):

Palloxin:
1.0.0 is already broken, webview doesnt work either

Palloxin:
I misunderstood the nature of the plugin.
Can u add an explaination for the common user in the error message that appears if one didnt insert credentials?

PR #1346, opened 2024-11-29
[Syosetu] Extract all web novel chapters listed in multiple pages

jacktheripper19:
- The previous version was only parsing pages from the first main page.
- Some web novels can have many chapters listed in multiple pages, so they need to be listed on the app as well.
- I could not find an API for listing all chapters from a single call. 
- This version will go to the last page to get the total number of chapters, then use the pattern of increasing the id number to iterate over all of them.
- Therefore it is slightly slow for loading two pages. It is still better than loading every page.


src/plugins/japanese/Syosetu.ts
```diff
@@ -1,3 +1,4 @@
+import * as cheerio from 'cheerio';
 import { load as loadCheerio } from 'cheerio';
```
nyagami, line 2: duplicatd  import

nyagami:
you should increase plugin version to notify new update to app

nyagami (approved):

nyagami (approved):

PR #1353, opened 2024-12-03
ReLibrary: Fix loading scraping+loading covers when a novel is parsed

Maix0:
This is a small fix that makes it possible to load covers from a parsed novel.
There is also a small fix that will trim `/` from the urls when storing them. this should be a utilty function to prevent code duplication in my optinion (my version is *not good* though); 

I don't know if this is a site specific thing, but all the `img` tags had the `src` attribute non-existant by default (aka when transmitted over the wire), but an attribute named `data-cfsrc` was present with the same value as the loaded website on a browser.

Currently my solution is to check both and prefer `img.src` if it exits

bumped version to 1.1.0, maybe a smaller bump should be used but I don't know :)

To note: the plugin does work normally now.

### DIFF
--- a/src/plugins/english/StellarRealm.ts
+++ b/src/plugins/english/StellarRealm.ts
@@ -1,0 +1,2978 @@
+import { fetchApi, fetchText } from '@libs/fetch';
+import { Plugin } from '@typings/plugin';
+import { Filters, FilterTypes } from '@libs/filterInputs';
+import { defaultCover } from '@libs/defaultCover';
+import { isUrlAbsolute } from '@libs/isAbsoluteUrl';
+
+interface SeriesChaptersResponse {
+  chapters: Chapter[];
+}
+
+interface Chapter {
+  id: number;
+  series_id: number;
+  slug: string;
+  title: null | string;
+  name: string;
+  number: number;
+  price: number;
+  unlocked_at: string | null;
+  index_at: string;
+  created_at: string;
+  group_id: null;
+  index: number;
+  is_premium: boolean;
+  group: null;
+}
+
+interface InertiaResponse<TProps> {
+  component: string;
+  props: TProps;
+  url: string;
+  version: string;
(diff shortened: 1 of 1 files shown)


src/plugins/english/relibrary.ts
```diff
@@ -164,7 +164,19 @@ class ReLibraryPlugin implements Plugin.PluginBase {
   name = 'Re:Library';
   icon = 'src/en/relibrary/icon.png';
   site = 'https://re-library.com';
-  version = '1.0.0';
+  version = '1.1.0';
+
+  imageRequestInit: Plugin.ImageRequestInit = {
+    headers: {
+      Referer: ReLibraryPlugin.site,
```
nyagami, line 171: Should be `this.site` instead? I dont think `site` is a static attribute

src/plugins/english/relibrary.ts
```diff
@@ -316,10 +344,10 @@ class ReLibraryPlugin implements Plugin.PluginBase {
           }
           chapters.push({
             name: loadedCheerio(chap_el).text(),
-            path: chap_path,
+            path: ReLibraryPlugin.trim_slashes(chap_path),
```
nyagami, line 347: should be `this.trim_slashes` instead

src/plugins/english/relibrary.ts
```diff
@@ -164,7 +164,19 @@ class ReLibraryPlugin implements Plugin.PluginBase {
...
+    headers: {
+      Referer: ReLibraryPlugin.site,
+    },
+  };
+
+  private static trim_slashes(url: string): string {
+    while (url.startsWith('/')) url = url.slice(1);
+    while (url.endsWith('/')) url = url.slice(0, url.length - 1);
+    return url;
+  }
```
nyagami, line 179: be careful, after this changes, novels in popular and search would be treated as new novels (though they were installed) because the path is not same as in the database

PR #1397, opened 2024-12-29
User plugin issue labeling

CD-Z:
This PR aims to create Labels for each plugin.
The reason for this is to force users to select the right plugin and get the issue labeled with the plugin id.
With this it will be possible to mark faulty plugins in the app with the help of an api call.
For example:
 - https://api.github.com/repos/LNReader/lnreader-plugins/issues?labels=Source%20Request,Theme:%20Madara

That way it will be easier to reduce duplicate issues and also provide a direct link from the app to the open issues of a plugin.

Current progress:
 [x] Dropdown with all plugin names (Can be viewed in my fork)
 [x] Dropdown for issue severity
 [x] Auto Issue labeling


.github/workflows/update_plugin_list.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Commit and Push Changes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add .github/ISSUE_TEMPLATE/report_issue.yml
+          git add .github/scripts/keys.json
+          if git diff --cached --quiet; then
+            echo "No changes to commit."
+          else
+            git commit -m "Update issue template with plugin dropdown options"
```
nyagami, line 33: this commit need `[skip ci]` prefix to not trigger publish pluginaction

.github/scripts/keys.json
```diff
@@ -0,0 +1 @@
...ranslations":"HasuTL","Español: LightNovelDaily":"lightnoveldaily","Español: MTL Novel (ES)":"mtlnovel-es","Español: Novelas Ligera":"novelasligera","Español: Oasis Translations":"oasistranslations","Español: Pancho Translations":"panchotranslations","Español: ReinoWuxia":"reinowuxia","Español: SkyNovels":"skynovels","Español: Traducciones Amistosas":"traducciones","Español: TuNovelaLigera":"tunovelaligera","Español: Yuuki Tls":"yuukitls","ไทย: Novel Lucky":"novel-lucky","ไทย: Novel PDF":"novelpdf","Türkçe: Araz Novel":"azraznovel","Türkçe: E-KİTAPLAR":"ekitaplar","Türkçe: EpikNovel":"epiknovel","Türkçe: Kodeks Library":"kodekslibrary","Türkçe: NABİ SCANS":"nabiscans","Türkçe: Namevt":"namevt","Türkçe: Novel oku":"noveloku","Türkçe: Novelokutr":"novelokutr","Türkçe: NovelTR":"noveltr","Türkçe: ThNovels":"thnovels","Türkçe: TurkceLightNovels":"turkcelightnovels","Türkçe: WebNovelOku":"webnoveloku","Українська: Смаколики":"smakolykytl","Tiếng Việt: Light Novel VN":"lightnovel.vn","Tiếng Việt: Hako":"ln.hako","Tiếng Việt: Nettruyen":"nettruyen","Tiếng Việt: Truyện Chữ":"truyenchu","Tiếng Việt: Truyen Conect":"truyenconect","Tiếng Việt: Truyện Full":"truyenfull","Multi: Komga":"komga"}
```
nyagami, line 1: is this file necessary?

nyagami:
I think the format `Plugin: <Plugin name>[Plugin id]` looks better.
eg: `Plugin: Hako[ln.hako]`
image

.github/workflows/update_plugin_list.yml
CD-Z: What do you mean exactly, I don't quite get it?

.github/scripts/keys.json
CD-Z: This file is automatically generated from the plugin.min.json file. It is for the ease of use and to reduce the requests. That way the labeling action is faster and it is also easier to implement.

CD-Z:
True

.github/workflows/update_plugin_list.yml
nyagami: Every commit in master branch trigger github action (Publish plugins).
In order to avoid it, add prefix `[skip ci]` in commit message

CD-Z:
I now changed the labels
image

CD-Z:
@nyagami I changed it now to a full cjs script instead of the bash script and the bot now creates a new branch with a pr.

CD-Z:
?

nyagami (approved):

PR #1453, opened 2025-01-23
Add fanfic search to webnovel

Soopyboo32:


src/plugins/english/webnovel.ts
```diff
@@ -214,10 +217,11 @@ class Webnovel implements Plugin.PluginBase {
   async searchNovels(
     searchTerm: string,
     pageNo: number,
+    type?: string,
```
nyagami, line 220: I dont think we have this param

Soopyboo32: ye we dont
its called above in the popular novels query cus i cba making a second function
i guess i could as a just in case for the future tho

Soopyboo32: did change

nyagami (approved):

PR #1532, opened 2025-02-22
Fix komga text color

Soopyboo32:


src/plugins/multi/komga.ts
```diff
@@ -214,6 +214,11 @@ class KomgaPlugin implements Plugin.PluginBase {
       }
     });
 
+    // Replace <a> tags with the text inside so its not blue
+    $('a').each((_, a) => {
+      $(a).replaceWith($(a).text());
```
nyagami, line 219: you can use plugin custom css instead

Soopyboo32: idk if that would be better tbh should links even exist in reader

cus u just click on them and the reader gets screwed up
(also most of the a tags in komga shouldent even be detected, in the response they are self closing but html doesent support self closing a tags so the tag just contains the entire rest of the chapter content

Soopyboo32: image
heres an example

Soopyboo32: *self closing tags should be valid btw as the html document is in XML mode that does support self closing tags (hence why the links arent there in browser) but the html parser used doesent seem to support it, even tho its in xml mode
```js
    const $ = parseHTML(htmlString, { xmlMode: true });
```

Soopyboo32: or maybe that does support but cus its also outputting in xml the reader view cant show it cus reader view doesent have the html doctype thing to put it in xml mode

Soopyboo32: *there are some properly formatted links sometimes like chapter titles sometimes link back to table of contents, and toc has links ect. but IMO having them be actuall links just makes the reader experience worse cus that is not how you should be navigating the novel (if it even works)

nyagami (approved):

PR #1555, opened 2025-03-08
src update: kolnovel

K1ngfish3r:
changed url from kolnovel.org -> kolbook.xyz
closes https://github.com/LNReader/lnreader-plugins/issues/1548
closes https://github.com/LNReader/lnreader-plugins/issues/1537

### DIFF
--- a/scripts/multisrc/lightnovelwp/sources.json
+++ b/scripts/multisrc/lightnovelwp/sources.json
@@ -10,13 +10,13 @@
   },
   {
     "id": "kolnovel",
-    "sourceSite": "https://kolnovel.org/",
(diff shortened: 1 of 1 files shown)


nyagami (approved):

TEKO00:
What about novels paradise?

K1ngfish3r:
> What about novels paradise?



Novels paradise works, but only a few people can access it using Webview. If you can access it via Webview without receiving a Error 403 Forbidden you should be fine.

I (we at the discord) have been unable to figure out why some devices can access whilst others cannot. You may use the following github issue to track this http://github.com/LNReader/lnreader-plugins/issues/1101

PR #1584, opened 2025-03-16
Turns out novelupdates dont work how i thought, so revert change

Soopyboo32:


src/plugins/english/novelupdates.ts
```diff
@@ -1004,7 +1004,7 @@ class NovelUpdates implements Plugin.PluginBase {
     );
```
Soopyboo32, line 1004: this is very wierd search logic, looks like its searching by the single longest word?

nyagami: weird tho.
The search I see in the website look different
image

image

Batorian: > weird tho. The search I see in the website look different image
> 
> image

I mean we could switch to API search.

PR #1591, opened 2025-03-23
Add new source

Hiirbaf:


nyagami:
Please implement PluginBase interface and use this [template](https://github.com/LNReader/lnreader-plugins/blob/master/docs/plugin-template.ts) to create new plugin

PR #1592, opened 2025-03-24
Feat: Chapter Sync

Batorian:


src/types/plugin.ts
```diff
@@ -87,6 +89,10 @@ export namespace Plugin {
      */
     parseNovel(novelPath: string): Promise<SourceNovel>;
     parseChapter(chapterPath: string): Promise<string>;
+    syncChapterStatus?(
+      novelPath: string,
+      chapterPath: string,
+    ): Promise<boolean>;
```
nyagami, line 95: I think we should name this more abstracty. Like `onChapterComplete`.
Have to consirder potental usecases to have a proper function name

Batorian: Good idea.

PR #1595, opened 2025-03-25
FIx mvlempyr

Soopyboo32:
:fire: (this took 8 hours to figure out :skull:)

### DIFF
--- a/proxy.ts
+++ b/proxy.ts
@@ -18,7 +18,13 @@ const settings: ServerSetting = {
     'sec-fetch-dest',
     'pragma',
   ],
-  disAllowResponseHeaders: ['link', 'set-cookie', 'set-cookie2'],
+  disAllowResponseHeaders: [
+    'link',
+    'set-cookie',
+    'set-cookie2',
+    'content-encoding',
+    'content-length',
+  ],
   useUserAgent: true,
 };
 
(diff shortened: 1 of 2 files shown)


proxy.ts
```diff
@@ -160,10 +160,12 @@ const proxyRequest: Connect.SimpleHandleFunction = (req, res) => {
...
-      .then(res2 => res2.text())
-      .then(res2 => {
-        res.statusCode = 200;
-        res.write(res2);
+      .then(async res2 => [res2, await res2.text()] as const)
+      .then(([res2, text]) => {
+        res.statusCode = res2.status;
+        res.setHeader('Content-Type', res2.headers.get('Content-Type'));
+        res.setHeader('X-Wp-Totalpages', res2.headers.get('X-Wp-Totalpages'));
+        res.write(text);
```
nyagami, line 168: we could return more headers like proxy
```js
  for (const _header in proxyRes.headers) {
    if (!settings.disAllowResponseHeaders.includes(_header)) {
      res.setHeader(_header, proxyRes.headers[_header] as string);
    }
  }
  ```

nyagami (approved):

proxy.ts
Soopyboo32: image
nvm that is prob missing some aswell

Soopyboo32: image
this is the error

Soopyboo32: ahh its cus its trying to decode the gzip from `content-encoding: gzip` header

nyagami (approved):

PR #1601, opened 2025-03-30
add scripts and description how to test plugins locally

eEQK:


package.json
```diff
@@ -11,8 +11,9 @@
     "generate": "node ./scripts/multisrc/generate.js",
     "host-linux": "chmod +x ./host.sh && ./host.sh",
     "host-windows": "powershell ./host.ps1",
+    "host-dev": "npx tsc --project tsconfig.production.json && npm run json && npx serve",
```
nyagami, line 14: Was server package added?

eEQK: npx command downloads it automatically, it doesn't have to be specified in `package.json` as a dependency

https://www.npmjs.com/package/serve

nyagami: So it would be a global package right? I dont think people want it. 
Or you could also serve json files via `vite` by moving them to public folder and change vite server host to 0.0.0.0

eEQK: I've updated it in a way you have described

nyagami: make sense, but could you add the serve package to devDependencies.

nyagami (approved):

PR #1609, opened 2025-04-02
69shu forgot pagination, added header

r1di:


src/plugins/chinese/69shu.ts
```diff
@@ -1,15 +1,29 @@
-import { load as parseHTML } from 'cheerio';
+import { load as parseHTML, CheerioAPI } from 'cheerio';
+import * as cheerio from 'cheerio';
```
nyagami, line 2: app doesnt support this import

nyagami (approved):

PR #1615, opened 2025-04-08
closes issue 1577, issue 1512, issue 1538,issue 1519

ElTonito974:
also issue 1264 needs tag closed because the sit is down so it is unaddable 
i signal that issue 1393 is not madara
finally issue 1463 needs to be closed because the site is present under a different name BoxNovel


Palloxin:
@ElTonito974 there are two problems with reaper scan:

1. the scraped summany scrapes the html tags
2. Defective webview opening, it's like a piece in the link is missing, giving no response
3. Extra: 
JbQEdu9Y
Can u polish the chapters a little bit? Maybe remove the "REAPPER SCANS" (useless), maybe also the title with the translator names IF they are in a tagg with special attribute

Issue #1611, opened 2025-04-05
Reaper plugin can't parse the content of old chapters


extra7:
### Plugin

English: Reaper Scans

### Severity

Missing chapter

### Steps to reproduce

1. Go to The 31st Piece Turns the Tables or Is It Bad That the Main Character’s a Roleplayer?
2. Open the first chapter
3. Find that it's content is empty

### Expected behavior

Find that The chapter content were correctly parsed and you're able to read it

### Actual behavior

Find an empty page and zero content in those first few chapters 

![Image](https://github.com/user-attachments/assets/a146ecc7-d829-40b3-826d-c67de99a8c18)

### LNReader version

Beta 2.0.0

### Android version

Android 10

### Other details

This problem doesn't manifest in the chapter that were uploaded not long ago aka the last bunch of chapters of the novels, it only happen in old chapters 

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[1.1.19](https://github.com/LNReader/lnreader/releases/latest)**.
- [x] I have updated all installed extensions.
- [x] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/LNReader/lnreader/issues/new/choose).

PR #1620, opened 2025-04-08
fix: reaperscans RSC parsing for multiline chapters (#1611)

eEQK:


Palloxin:
I didnt test this pr changes, but here the 2-4 problems with the currently avaiable plugin version on the repo:

1. the summaries have the html tags in them
2. Defective webview opening, it's like a piece in the link is missing, giving no response
3. Extra: 
JbQEdu9Y
Can u polish the chapters a little bit? Maybe remove the "REAPPER SCANS" (useless), maybe also the title with the translator names IF they are in a tagg with special attribute

4. certain covers do not load, just open reaper scans main screen on the phone and see. They are just there, not loading, some loads after some time tho.

nyagami (approved):

PR #1624, opened 2025-04-10
new plugin novel543(chinese)

r1di:
-search not working well
--search over the novelid works
--normal search might not work because of cloudflare...


src/plugins/chinese/novel543.ts
```diff
@@ -0,0 +1,326 @@
...
+        });
+      }
+    });
+
+    return novels;
+  }
+
+  async fetchImage(url: string): Promise<string | undefined> {
+    return await fetchFile(url);
+  }
```
nyagami, line 323: there's no fetchImage anymore, use `imageRequestInit` instead

nyagami (approved):

PR #1654, opened 2025-04-25
Freewebnovel: use htmlparser2

K1ngfish3r:
-> fixes malformed tags which cause them to go missing in the text (hopefully)
-> improve load times for novels

technically an experimental release since the malformed tag catching code might end up breaking the chapter


package-lock.json
nyagami, line 1: why there're changes in package-lock.json but nothing in package.json

K1ngfish3r: i `rm -rf node_modules` and `npm i`
suddenly it got more

nyagami (approved):

PR #1683, opened 2025-05-16
Ranobelib: Fix cant parse page

K1ngfish3r:
Fixes everything i think, cant test cause rider missing

### DIFF
--- a/src/plugins/russian/ranobelib.ts
+++ b/src/plugins/russian/ranobelib.ts
@@ -17,7 +17,7 @@ class RLIB implements Plugin.PluginBase {
   id = 'RLIB';
   name = 'RanobeLib';
   site = 'https://ranobelib.me';
-  apiSite = 'https://api.mangalib.me/api/manga/';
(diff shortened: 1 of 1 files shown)


nyagami (approved):

VectorNEON:
After updating the library, the chapters began to double and the translator was added everywhere "1" which is the reason for the doubling of the chapters
image

K1ngfish3r:
> After updating the library, the chapters began to double and the translator was added everywhere "1" which is the reason for the doubling of the chapters 

can you provide a novel name? @Rider21 is probably needed

VectorNEON:
> > After updating the library, the chapters began to double and the translator was added everywhere "1" which is the reason for the doubling of the chapters
> 
> can you provide a novel name? @Rider21 is probably needed

In principle, any short story where one translator, for example, can be checked - Низкоуровневый персонаж Томозаки-кун - https://ranobelib.me/ru/book/24527--novel?ui=2555856

PR #1750, opened 2025-07-07
Fix mvlempyr again following website recode

Soopyboo32:
This should also make loading the list of novels like 10x faster cus it can be done in 1 api call now :fire:


Soopyboo32:
image
sheeshers

nyagami (approved):

PR #1768, opened 2025-07-24
added StellarRealm source

rafi1212122:
Would've been better if i could asynchronously load filters, but this works ig


Batorian:
Hey, I really like your clean implementation of the Stellar Realm plugin. Could you maybe take a look at the Stellar Realm implementation in the Novel Updates plugin when you have time? I've been kinda stuck as to why it's throwing network errors when trying to open chapters.

If you want to contact me you can easily do so in the 'plugin-beta' Discord channel.

PR #1776, opened 2025-08-05
Add KDT Novels plugin

ripdog:
Adds a source for https://kdtnovels.com. 

It's vibe coded, but I monitored it closely and have tested all functions, on the web UI and on a real device.

Happy to make changes if the code style isn't desirable. I know it's a bit wordy.


src/plugins/english/kdtnovels.ts
```diff
@@ -0,0 +1,403 @@
...
+      console.error('Error searching novels:', error);
+      return [];
+    }
+  }
+
+  async parseNovel(novelPath: string): Promise<Plugin.SourceNovel> {
+    // Construct full URL from novel path
+    const novelUrl = novelPath.startsWith('http')
+      ? novelPath
+      : `${this.site}/${novelPath.replace(/^\//, '')}`;
```
nyagami, line 158: if novel/chapter paths contain domain, users would have duplicated chapters because app use paths for identifying novel/chapter from remote source

ripdog: Okay, following the example of other plugins, I've removed all the path normalisation.

nyagami (approved):

Issue #891, opened 2024-04-17
Blank chapter


MadameVerre:
### Source information

SakuraNovel

### Steps to reproduce

Open the source and search for the title

### Expected behavior

When in the title, chapters do not appear 

### Actual behavior

I can read normally in the webview

### LNReader version

1.1.19

### Android version

6

### Other details

Chapters do not appear after opening the title, I can read normally in webview

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[1.1.19](https://github.com/LNReader/lnreader/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/LNReader/lnreader/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

nyagami:
https://discord.com/channels/835746409357246465/1224743197843132469/1224782544894165003
The site doesnt want to be scraped lol.

PR #1832, opened 2025-10-04
Remove SakuraNovel plugin

error7404:


nyagami:
Is it dead?

error7404:
No but the site doesn't want to be scraped ref: #891

Palloxin:
what website likes to be scraped

PR #1857, opened 2025-10-24
Fix Story seedling

Soopyboo32:


plugins/index.ts
nyagami: Wait, why do we push this file to github. It should be generated by local dev and ignored from git

Soopyboo32: good question, idk

Soopyboo32: looks like gitignore doesent filter it

PR #2166, opened 2026-05-03
fix(novelbuddy): wrap summary in <div> so cheerio always HTML-parses it

ig4e:
## Bug

`parseNovel` calls `$(initialManga.summary || '')`. Cheerio's `$()` only takes the HTML branch when the input starts with `<` — otherwise it parses the input as a **CSS selector**. The novelbuddy API returns `initialManga.summary` as plain text for many novels (e.g. *Trash of the Count's Family*, where the summary is `"When I opened my eyes, I was inside a novel."`), so the selector parser hits the first `.` from punctuation and throws:

```
Error: Expected name, found .
```

The whole `parseNovel` call fails and the novel page can't be opened.

## Repro

1. Open NovelBuddy
2. Open *Trash of the Count's Family*
3. Details fail with `Expected name, found .`

## Fix

Wrap the summary in a `<div>` before passing to `$()`. Plain text becomes `<div>plain text</div>` (parses as HTML, `.text()` returns the original string), HTML stays HTML, and the existing `.find('br')` / `.find('p')` transforms still work either way.

Bumps version `2.1.0` → `2.1.1`.

### DIFF
--- a/plugins/english/novelbuddy.ts
+++ b/plugins/english/novelbuddy.ts
@@ -9,7 +9,7 @@ class NovelBuddy implements Plugin.PluginBase {
   name = 'NovelBuddy';
   site = 'https://novelbuddy.com/';
   api = 'https://api.novelbuddy.com/';
-  version = '2.1.0';
+  version = '2.1.1';
   icon = 'src/en/novelbuddy/icon.png';
 
   parseNovels(body: Response): Plugin.NovelItem[] {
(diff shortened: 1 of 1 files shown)


K1ngfish3r:
nice catch, i had resorted to using the summary in the novel page instead in my lint pr

ig4e:
Thanks @K1ngfish3r, sadly the ln reader app doesn't expose errors just keeps loading indefinitely. i caught this while working on a js-kotlin birdge for my app, at first i thought it was on my end

rajarsheechatterjee (approved):

PR #2176, opened 2026-05-06
Add I'm Reading A Book plugin (Spanish)

anehemias6-star:
Adds a new Spanish plugin for imreadingabook.com, a site that translates Korean, Chinese and Japanese light novels to Spanish.

- popularNovels: parses /nuestros-proyectos/ for the full novel list
- parseNovel: extracts cover, synopsis, status and chapter list from table
- parseChapter: extracts chapter text from WordPress entry-content
- searchNovels: uses WordPress search (?s=term)


K1ngfish3r (changes requested):
imreadingabook icon is going to an incorrect location, it should under public/static/

remove PanchoTranslations[madara]
it is a source that is already being generated inside madara multisrc
if you are going to move out from there, 
1. remove from /plugins/multisrc/madara/sources.json
2. move icon from multisrc into src
3. create a separate plugin and test correctly
4. bump version number above the madara version

PR #2197, opened 2026-05-17
fix(proxy): propagate origin status code and handle 304 correctly

7ui77:
This PR fixes a critical issue in the web debugger's proxy server that was causing 'Unexpected end of JSON input' errors when debugging plugins.

### Issue
The proxy was hardcoding the response status code to `200` and failing to propagate the actual status from the target server. When a server returned `304 Not Modified` (with an empty body), the proxy would still send `200` but with the empty body. The plugin, expecting valid JSON for a `200` response, would then fail with a JSON parsing error.

### Fix
- Removed hardcoded `res.statusCode = 200`.
- Implemented status code propagation in the `proxyRes` event handler.
- Added explicit handling for `304` responses to simply end the response without attempting to process a body.
- Improved decompression handling for various encodings (Brotli, Gzip, Zstd).

### Verification
Verified using the Web Debugger with the NovelBuddy plugin. Subsequent "Fetch" requests that return `304` now correctly trigger the browser's cache usage instead of causing JSON parsing errors in the plugin.


proxy.ts
```diff
@@ -105,78 +90,41 @@ const proxyHandlerMiddle: Connect.NextHandleFunction = (req, res) => {
...
-  console.log('\x1b[36m', '----------------');
-  console.log(
-    `Making proxy request - at ${new Date().toLocaleTimeString()}
-  url: ${_url.href}
-  headers:`,
-  );
-  Object.entries(req.headers).forEach(([name, value]) => {
-    console.log('\t', '\x1b[32m', name + ':', '\x1b[37m', value);
-  });
-  console.log('\x1b[36m', '----------------');
```
K1ngfish3r, line 125: why delet

7ui77: prevent sensitive information (like Auth tokens and Cookies) from being leaked in the logs. I'll add the logs back.

proxy.ts
```diff
@@ -211,94 +191,84 @@ const proxyRequest: Connect.SimpleHandleFunction = (req, res) => {
...
+  proxyRes.on('end', () => {
+    try {
+      let buffer = Buffer.concat(chunks);
+      if (buffer.length > 0) {
+        if (contentEncoding.includes('br'))
+          buffer = brotliDecompressSync(buffer);
+        else if (contentEncoding.includes('gzip')) buffer = gunzipSync(buffer);
+        else if (contentEncoding.includes('zstd'))
+          buffer = zstdDecompressSync(buffer);
+        res.write(buffer);
```
K1ngfish3r: fix type errors for buffer please
`"Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Buffer<ArrayBuffer>'.
Type 'ArrayBufferLike' is not assignable to type 'ArrayBuffer'.
Type 'SharedArrayBuffer' is not assignable to type 'ArrayBuffer'.
Types of property '[Symbol.toStringTag]' are incompatible.
Type '\"SharedArrayBuffer\"' is not assignable to type '\"ArrayBuffer\"'.",
}]`
am testing now

7ui77 (commented):
@K1ngfish3r mismatch between ArrayBufferLike (which includes SharedArrayBuffer).  Try this:
```Typescript
const compressedBuffer = Buffer.concat(chunks);
if (compressedBuffer.length > 0) {
    let decompressedBuffer: Buffer;
    if (contentEncoding.includes('br')) {
        decompressedBuffer = brotliDecompressSync(compressedBuffer) as Buffer;
    } else if (contentEncoding.includes('gzip')) {
        decompressedBuffer = gunzipSync(compressedBuffer) as Buffer;
    } else if (contentEncoding.includes('zstd')) {
        decompressedBuffer = zstdDecompressSync(compressedBuffer) as Buffer;
    } else {
        decompressedBuffer = compressedBuffer;
    }
    res.write(decompressedBuffer);
}
```

proxy.ts
```diff
@@ -120,20 +117,25 @@ const proxyRequest: Connect.SimpleHandleFunction = (req, res) => {
...
-    console.log('\t', '\x1b[32m', name + ':', '\x1b[37m', value);
+    // Redact sensitive headers in logs
+    const isSensitive = ['cookie', 'authorization'].includes(name.toLowerCase());
+    console.log(
+      '\t',
+      '\x1b[32m',
+      name + ':',
+      '\x1b[37m',
+      isSensitive ? '[REDACTED]' : value,
+    );
```
K1ngfish3r: Image

It's dev test for a reason, revert this

proxy.ts
```diff
@@ -211,94 +199,89 @@ const proxyRequest: Connect.SimpleHandleFunction = (req, res) => {
...
-    for (const _header in proxyRes.headers) {
-      if (!settings.disAllowResponseHeaders.includes(_header)) {
-        res.setHeader(_header, proxyRes.headers[_header] as string);
-      }
+  // Propagate headers but filter restricted ones
+  Object.keys(proxyRes.headers).forEach(key => {
+    if (
+      !settings.disAllowResponseHeaders.includes(key) &&
+      key !== 'content-encoding' &&
+      key !== 'content-length'
```
K1ngfish3r, line 240: any reason why these two keys were specified here?

K1ngfish3r (changes requested):
run prettier here too

proxy.ts
7ui77: client is now raw data. If you keep 'content-encoding: gzip', the browser will attempt to decompress it again, causing an error.
The data size after decompression will definitely change compared to the original. If you keep the old 'content-length', the browser will either freeze—waiting to receive the full original byte count—or throw an 'Unexpected end of data' error.

proxy.ts
7ui77: ok 👌,  Okay, I'll return it.

K1ngfish3r (approved):
this fixes covers on a few sources

PR #2247, opened 2026-06-11
fix(novelbin): fix missing summary and remove NB Reader advertisements

7ui77:
This PR resolves two issues with the NovelBin plugin:
1. Fixes the empty novel summary/description caused by the multiple class names on the description wrapper `div` (`desc-text desc-text-collapsed`).
2. Adds a customJs script to automatically strip the "NB Reader" mobile app promotion/advertisements from chapter contents.

### Changes:
- Added case for `'desc-text desc-text-collapsed'` in the `parseNovel` method of the `readnovelfull` multisrc template.
- Configured a simplified `customJs` option for `novelbin` inside `sources.json` to remove elements with classes matching `app-promo`.
- Bumped NovelBin version increment (`versionIncrements: 1`).

### DIFF
--- a/plugins/multisrc/readnovelfull/sources.json
+++ b/plugins/multisrc/readnovelfull/sources.json
@@ -73,7 +73,8 @@
     "sourceName": "Novel Bin",
     "options": {
       "latestPage": "sort/latest",
-      "searchPage": "search"
+      "searchPage": "search",
+      "versionIncrements": 1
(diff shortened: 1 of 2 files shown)


K1ngfish3r:
need to bump version number in template

7ui77:
> need to bump version number in template

Done!

PR #2317, opened 2026-07-17
refactor(lnori): use a pageCache rather than fetching the same page

K1ngfish3r:
#### Checklist

- [x] Reference related issues in the PR body (e.g. Closes #xyz)

No relevant issues

- use a pool to send 5 requests instead of 30 at the same time
- cache the page so the page doesnt have to be downloaded over and over during chapter downloads

### DIFF
--- a/plugins/english/lnori.ts
+++ b/plugins/english/lnori.ts
@@ -3,6 +3,7 @@ import { storage } from '@libs/storage';
 import { Plugin } from '@/types/plugin';
 import { Filters, FilterTypes } from '@libs/filterInputs';
 import { load as parseHTML } from 'cheerio';
+import { Parser } from 'htmlparser2';
 import { defaultCover } from '@libs/defaultCover';
 
 class LnorisPlugin implements Plugin.PluginBase {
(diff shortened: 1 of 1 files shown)


MonomoySama:
For Lnori, i think the original is good enough, although it takes 2min+ for big entries, like Rezero, Mushoku tensei etc. at least they open fully and will later stay opened in cache. The problem is just that one volume will always be missing for some reason and some novels like classroom elite and reincarnated as slime are "can't parse page". So rather than increasing loading speed, i think fixing that missing volume problem is good enough, as for parse page, i think it's related to that missing volume problem. And it's not like the site updates new volumes every day that we gotta update entries everyday.

MonomoySama:
Idk why and how, but it's working fully fine now.... Wtf? Cant parse page novels, fixed. Missing volume, fixed..... Just that the loading speed is same. So no one fixed anything and it got fixed itself? Hmm... Hope it's not temporary. Well then, you can focus on making a novel phoenxi plugin, so i can migrate there from novelfire

MonomoySama:
nevermind, was indeed a temporary fix. now giving same problems again. see this #2334 maybe the problem is how it's sending the fetch requests

K1ngfish3r:
I am assuming the number of requests become too many due to this plugin not caching anything

MonomoySama:
> I am assuming the number of requests become too many due to this plugin not caching anything

Well, last time you almost fixed it, loading time became instant, all volumes were there too, but only for few novels, rest were all inf loading, remember?

K1ngfish3r:
ie: other people using the current plugin, unrelated to this PR

PR #2320, opened 2026-07-17
fix(FreeWebNovel): fetch all chapter pages when paginated

Dakrataz:
## Summary
FreeWebNovel paginates its chapter list (40 chapters per page). The plugin was only parsing page 1, so novels with more than 40 chapters were missing the rest.

This fix:
- Reads the `data-total-page` attribute from the `#indexListPage` div
- Fetches remaining pages via the site's AJAX endpoint (`?ajax=chapters&page=N`)
- Combines all chapters into one list using cheerio

#### Checklist


K1ngfish3r:
Pagination will end up spamming the website, especially for novels with 1000+ chaps (1000/40 => 25 requests just for chapter listing). We shouldn't spam if we don't need to

PR #2323, opened 2026-07-18
feat(arabic): add KolNovel plugin

reda57939abd-hash:
Add KolNovel plugin for Arabic language.

This plugin handles the KolNovel site and uses proper parsing logic with Cheerio, bypassing copy-protection attributes.


K1ngfish3r:
this plugin already exists, recommend fixing the customJs in the sources.json for kolnovel and bump the ver number unless there is a particular need for a completely different implementation

PR #2328, opened 2026-07-18
Feat rewayatfans

hamedhani1998:
#### Checklist

- [ ] Update version code if an existing plugin was modified
- [ ] Test changes in Plugin Playground or the app


rajarsheechatterjee (changes requested):
*This review was generated by Codex.*

Codex found the following structural issues that should be addressed before merging:

* `plugins/arabic/rewayatfans.ts:5-13,123-124`: This change removes `content` from `WPPage`, but `parseChapter` still reads `arr[0].content.rendered`. A strict type-check now reports TS2339 on both accesses. Restore `content: { rendered: string }` (or use a separate response type that includes it) so the chapter response matches the code that consumes it.

PR #2355, opened 2026-07-26
fix(chireads): adapt to new site theme and refresh tag filters

elfing44:
- Full fix of novel and chapter fetching after the site redesign (new `refresh-*` selectors, shared `parseCards`, categories fetched in parallel, chapter date via regex)
- Refresh the tag filter against the live tags: drop empty tags
  (`adulte`, `croissance-acceleree`), add `Drame` and `Magie`, sort
  alphabetically

#### Checklist

### DIFF
--- a/plugins/french/chireads.ts
+++ b/plugins/french/chireads.ts
@@ -6,12 +6,32 @@ import dayjs from 'dayjs';
 import { defaultCover } from '@libs/defaultCover';
 import { NovelStatus } from '@libs/novelStatus';
 
+type CardSelectors = {
+  item: string;
+  titleLink: string;
+  cover: string;
+};
+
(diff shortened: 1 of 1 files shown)


K1ngfish3r:
search seems to fetch 8 pages, 4 of which dont exist (`https://chireads.com/category/original/` doesn't have pages afaik)
might be easier to cache the results

K1ngfish3r:
how about using `chireads.com/search?x=0&y=0&name=${encodeURIComponent(searchTerm)}`? apparently if you press enter it always defaults to 0,0 and you can change these values depending on where you click on the search button img

UPDATE: apparently you can search yourself into chapters, ignore this suggestion

PR #2363, opened 2026-07-30
feat: Add Skythewood translations

creeperkafasi:
This PR adds support for [Skythewood Translations](https://skythewood.blogspot.com/)

I didn't really check every single novel that gets scraped, but i know some novels don't show up and some might not even work. "Too many losing heroines" works, i tested it.
Scraping the website was messy because it doesn't seem to have much of a rigid structure to it, especially with older projects.

better than nothing i guess `~('v')~`

*I'd appreciate it if you could squash these commits before merging*

#### Checklist

- [ ] Update version code if an existing plugin was modified


plugins/english/skythewood.ts
```diff
@@ -0,0 +1,243 @@
...
+import { CheerioAPI, load as parseHTML } from 'cheerio';
+import { Element } from 'domhandler';
+import { Filters } from '@/types/filters';
+import { Plugin } from '@/types/plugin';
+
+class SkyTheWood implements Plugin.PluginBase {
+  id = 'skythewoodtranslations';
+  name = 'Skythewood Translations';
+  site = 'https://skythewood.blogspot.com';
+  icon = 'favicon.ico';
```
K1ngfish3r: you'll have to add an icon.png (96x96 px) to `public/static/src/en/skythewood` and change `favicon.ico` to reflect the path

plugins/english/skythewood.ts
```diff
@@ -0,0 +1,243 @@
...
+  version = '1.0.0';
+
+  async popularNovels(
+    pageNo: number,
+    _options: Plugin.PopularNovelsOptions<Filters>,
```
K1ngfish3r:
```suggestion
```
can simply remove this and the filters import since they are unused

plugins/english/skythewood.ts
```diff
@@ -0,0 +1,243 @@
...
+    type SkyProjects = {
+      names: string[][];
+      novels: {
+        name: string;
+        href: string;
+        cover: string | undefined;
+      }[];
+    };
+
+    let doneProjects: SkyProjects = await this.getDoneProjects();
```
K1ngfish3r: please run `npm run lint -- --fix` , this will stop linter from complaining

K1ngfish3r (changes requested):
missing metadata for now is ok, can try fixing the rest of the novels later
fix these issues and i squash

K1ngfish3r:
technically, we use `this.site` and keep paths separate so that if the site changes domain the users' chapter lists won't be affected when the URL is updated, but looking at the fact that this is a blogspot site, we can probably keep full paths, changing the website domain would probably massively update the layout to the point where a massive refactor would be required. but your call

creeperkafasi:
> technically, we use `this.site` and keep paths separate so that if the site changes domain the users' chapter lists won't be affected when the URL is updated, but looking at the fact that this is a blogspot site, we can probably keep full paths, changing the website domain would probably massively update the layout to the point where a massive refactor would be required. but your call

I agree with you on the layout change. But they don't seem to have changed their domain since 2014 which is the earliest instance of the blog being used as a tl site on archive.org, so i think it's safe to assume they aren't going anywhere or changing their layout

creeperkafasi:
Fixed and tested with your in editor suggestions applied, thanks

K1ngfish3r:
> > technically, we use `this.site` and keep paths separate so that if the site changes domain the users' chapter lists won't be affected when the URL is updated, but looking at the fact that this is a blogspot site, we can probably keep full paths, changing the website domain would probably massively update the layout to the point where a massive refactor would be required. but your call
> 
> I agree with you on the layout change. But they don't seem to have changed their domain since 2014 which is the earliest instance of the blog being used as a tl site on archive.org, so i think it's safe to assume they aren't going anywhere or changing their layout

oops, i checked the epub logic, it uses resolveUrl which would prepend the this.site to the url... i think...
might need refactor

K1ngfish3r:
> oops, i checked the epub logic, it uses resolveUrl which would prepend the this.site to the url... i think... might need refactor

ignore this crap, epubs are already downloaded, can skip refactor

PR #2520, opened 2026-09-13
Push mknov upstream

hamedhani1998:
#### Checklist

- [x] Reference related issues in the PR body (e.g. Closes #xyz)
- [x] Commit messages follow `type(scope): description` (e.g. `feat(<generator>): add new source`)


.js/src/plugins/arabic/mknov.js
rajarsheechatterjee, line 1: Could you remove this?

hamedhani1998: Hi — this `.js` is the compiled output of `plugins/arabic/mknov.ts`. I only committed it (as a separate "chore" commit) so the plugin could be loaded and tested in the app immediately, without pulling from a build step first.

The file itself is not required: the repo `.gitignore` already lists `.js`, and none of the other plugins under `plugins/arabic/` commit compiled output — the build generates it from the `.ts`. I will remove it from the PR so it only ships the TypeScript source.

hamedhani1998: Hi — this `.js` is the compiled output of `plugins/arabic/mknov.ts`. I only committed it (as a separate "chore" commit) so the plugin could be loaded and tested in the app immediately, without pulling from a build step first.

The file itself is not required: the repo `.gitignore` already lists `.js`, and none of the other plugins under `plugins/arabic/` commit compiled output — the build generates it from the `.ts`. I will remove it from the PR so it only ships the TypeScript source.

hamedhani1998: Done — the compiled `.js` is removed. The PR now ships only `plugins/arabic/mknov.ts` (+ the icon), same as the other plugins.

PR #2527, opened 2026-09-16
fix(english/novelight): serve 50-chapter pages from the site's own page ranges

### LINKED ISSUE #2526: Pagination Issue - inconsistent & missing chapters
### Plugin

English: Novelight

### Severity

Wrong Formatting

### Steps To Reproduce

Open any novel from plugin with more than 50 chapters.

### Expected Behavior

Each page should only have 50 chapters by default. After 50 chapters, there should be a new page with the following chapters.

### Actual Behavior

Chapters continue appearing on same page, even after refreshing.

### LNReader Version

2.1.3

### Plugin Version

1.1.5

### Android Version

Android 14

### Other Details

This happened ever since I upgraded to version 2.1.3. My library migrated successfully, but this plugin started having this issue. This causes the number of chapters on each page to be inconsistent, and sometimes chapters are missing.

For example, with the novel "Best Friend X Swap", the first page has 44 chapters, the second page has 8 chapters, and from the third page onwards they have 50 chapters.

https://github.com/user-attachments/assets/a76aa2e0-4f41-4ee0-bcbd-c7acd8ffbd13

Another example is with the novel "Our Guild Princess is Good at Everything", where the first page has only 1 chapter, the second page has 50 chapters, and the third page has 55 chapters.

https://github.com/user-attachments/assets/20d4b213-db93-4c0c-9d61-23bd3ac3ec01

Last example I have is with the novel "Regression Guildines for the Supporting Character", where the first page has 27 chapters, the second to eighth pages have 50 chapters each, and the ninth page has 71 chapters. However, the eighth page ends with chapter 374, while the ninth page starts with chapter 376. Chapter 375 is entirely missing. It's viewable on webview, but it doesn't appear in-app no matter how many times I refresh.

https://github.com/user-attachments/assets/58cb0c3b-5477-4028-be3d-d828870fd6d6

Sorry for the lengthy post! I didn't want to make an issue report for every individual novel since it definitely seems like a plugin issue.

### Acknowledgements

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have written a clear and descriptive title.
- [x] I have updated the app to the latest version **[2.0.0](https://github.com/LNReader/lnreader/releases/tag/v2.0.0-beta.3)**.
- [x] I have updated all installed plugins to the latest version.
- [x] I understand that app-specific issues should be reported in the [app repository](https://github.com/LNReader/lnreader/issues/new/choose).

RibatTRW:
## Fix

Closes #2526. Long Novelight chapter lists split into wrong 50-chapter pages: page 1 came out partial (`[2, 50, 50, 50, 50]` on a 202-chapter novel) because `parseNovel` and `parsePage` derived the page count two different ways, and the site's ajax endpoint silently clamps out-of-range pages to its last page, turning any disagreement into duplicated chapters.

- `parseNovel` and `parsePage` now share one partition derived from the site's own page ranges: `totalPages = ceil(totalChapters / 50)`, each page sliced to exactly its 50-chapter chunk with the partial page last. Chunks anchor at the oldest chapter, so new chapters only extend the last page. Same approach as `plugins/english/mynovels.ts`.
- Every request sends browser-like headers (`User-Agent`, `Referer`, `Accept-Language`): the plugin sent no `User-Agent`, and Cloudflare answers UA-less requests from distrusted networks with a bot-check page that parses to zero novels (this failed the live check's `popularNovels` from CI).
- Refused responses throw carrying the HTTP status, so a runner-side block reports INCONCLUSIVE instead of a false empty result.
- Version `1.1.5` → `1.1.8`. Chapter name, URL, date parsing and `hideLocked` behavior are unchanged.

## Verification

Live against the real site with the bundled plugin: all three sampled novels paginate exactly (`[50, 50, 50, 50, 2]`, `[50, 50, 6]`, `[50 × 8, 49]`), 202/202, 106/106 and 449/449 chapters served with zero missing or duplicated. `check:plugin` passes all four steps (30 popular, 1 search, 50-chapter novel, ~12k-char chapter). `build:compile` clean; the file is clean under lint and format (the 2 lint errors and 12 format findings in full runs are pre-existing in unrelated files).

Caveats: the exact "chapter 375 missing" symptom was not reproducible on the live site (the two old derivations currently agree there); no tested chapter is detected as locked, so `hideLocked` is a no-op there. Because the app caches chapters per page, a novel must be refreshed or re-added after updating. AI-authored change, no human review.


RibatTRW:
works also

RibatTRW:
Pushed v1.1.8 for this head: every request now goes through one helper that treats a refused response as a site block (reported INCONCLUSIVE per docs/testing.md) instead of parsing a block page into an empty catalog. Local `npm run check:plugin -- plugins/english/novelight.ts` passes fully (30 popular, search, 50 chapters, chapter text). The check run for this head is waiting on approval — could you approve it so the check can report?

RibatTRW:
The live check should work now on this head too. The previous failure class is a runner-side block being parsed into an empty result and reported as a plugin failure; every request here now carries browser-like headers and treats a refused response as a site block, which docs/testing.md defines as INCONCLUSIVE rather than a plugin FAIL. Verified against the live site (popular 30, search, 50 chapters per page, chapter text) and driven through the plugin playground in the app UI. The check run is waiting on approval — could you approve it so it can report?

PR #2528, opened 2026-09-16
fix(indonesian/indowebnovel): restore chapter list and chapter text

### LINKED ISSUE #2524: IndoWebNovel
### Plugin

Bahasa Indonesia: IndoWebNovel

### Severity

Can't Load Novels

### Steps To Reproduce

IndoWebNovel I've added a novel, but the chapters aren't showing up. 

https://github.com/user-attachments/assets/ec123642-ba6a-4a09-b13d-f822d597b60d

### Expected Behavior

https://github.com/user-attachments/assets/84af6953-e524-4174-abcb-3d272a224f81

### Actual Behavior

This error has been around since 1-1-2026 I think 

### LNReader Version

2.14

### Plugin Version

1.2.3

### Android Version

Android 12 

### Other Details

Image
It's like that and the chapter doesn't appear. 

### Acknowledgements

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have written a clear and descriptive title.
- [x] I have updated the app to the latest version **[2.0.0](https://github.com/LNReader/lnreader/releases/tag/v2.0.0-beta.3)**.
- [x] I have updated all installed plugins to the latest version.
- [x] I understand that app-specific issues should be reported in the [app repository](https://github.com/LNReader/lnreader/issues/new/choose).

RibatTRW:
## Fix

Closes #2524. IndoWebNovel novels came back with an empty chapter list and empty chapter text after a site markup change: the list container is now `ul.series-chapterlists` (plural) and the body no longer lives in `.adsads`, so both old selectors matched nothing.

- Chapter list reads `.series-chapterlists li` (site still newest-first, so the existing reverse to oldest-first stays); chapter names come from the entry's non-date span so the release date no longer runs into the title.
- Chapter body anchors on the stable inner `main #content` instead of a class name: the wrapper carries a long randomized-looking token, so nothing depends on its value.
- Listing accepts all three card layouts (`.flexbox2-item`, `.flexbox3-item`, `.popular .flexbox-item`) so a variant layout never parses to zero novels; chapter entries populate `releaseTime` from `span.date` where present.
- All requests send browser-like headers and throw carrying the HTTP status on a refused response, so blocks report INCONCLUSIVE instead of false empties.
- Version `1.2.4` → `1.3.1`. Title, cover, author, status, summary, genres, and path handling are unchanged.

## Verification

`check:plugin` passes all four steps live (20 popular, 1 search result, 497-chapter novel, 15282-char chapter). Also verified: a second 400-chapter novel, three chapter bodies with real prose and no boilerplate, and metadata populated on both novels. `build:compile` clean; the file is clean under eslint and prettier (remaining repo-wide findings are pre-existing in untouched files).

Caveat: the wrapper token never rotated across samples (same value in 4 chapters, repeat and cache-busted requests), so rotation itself was not reproduced — but the fix does not depend on the token. AI-authored change, no human review.


RibatTRW:
it works

plugins/indonesian/indowebnovel.ts
```diff
@@ -88,8 +88,16 @@ class IndoWebNovel implements Plugin.PluginBase {
 
     const chapters: Plugin.ChapterItem[] = [];
 
-    loadedCheerio('.series-chapterlist li').each((i, el) => {
-      const chapterName = loadedCheerio(el).find('a').text().trim();
+    loadedCheerio('.series-chapterlists li').each((i, el) => {
+      // The list entry also holds a `span.date`; drop it so the chapter name
+      // does not end with the release date.
```
rajarsheechatterjee, line 124: If the release date can be parsed, could we populate it for each chapter?

RibatTRW: Done,each chapter now populates releaseTime from the list entry's span.date (e.g. "October 23, 2025"), validated with Date.parse and stored as ISO; entries without a parsable date leave it unset. Verified against the live site: parseNovel returns 497 chapters all carrying releaseTime, and the plugin live check passes 4/4. (Implemented with AI assistance, verified locally; upstream CI on the fork branch is awaiting maintainer approval to run.)

RibatTRW:
Pushed 32176236 for this head: browser-like request headers, and a refused response now reports as a site block (INCONCLUSIVE per docs/testing.md) instead of an empty catalog; the chapter release dates asked about in the review thread are populated (497 chapters carry releaseTime, verified against the live site). Local live check passes 4/4. The check run for this head is waiting on approval — could you approve it so the check can report?

RibatTRW:
@rajarsheechatterjee the live check should work now. What failed was not the plugin: the runner was served a block page instead of the listing, and because `fetchApi` never throws on a non-2xx response the check parsed that page into an empty result and reported it as "Returned no novels". The plugin now sends browser-like request headers and treats a refused response as a site block, which docs/testing.md defines as INCONCLUSIVE rather than a plugin failure. Verified here against the live site (popular 20, novel 497 chapters with dates, chapter text 15k chars, search results) and additionally driven through the plugin playground in the app UI, where every view rendered real content. We cannot start the workflow from the fork without approval — could you approve the check run so it can report?

PR #2529, opened 2026-09-16
fix(ar): empty chapter bodies and Riwyat novel metadata (Galaxy Novels, Riwyat, Kol Novel, Free Kol Novel)

RibatTRW:
Closes #2522, #2568, #2486, #2483, #2505, #2465, #2501, #2370, #1566, #2442

Opening a chapter in these sources showed "no content" while the cover, description and chapter list still loaded. Each source was broken for a different reason; four are fixed, and the fifth could not be checked at all and is reported as not verified rather than claimed.

| Source | Cause | Status |
| --- | --- | --- |
| Galaxy Novels | chapter links were built from a WordPress post id (`…/chapter-<post-id>/`, which 404s), and the JSON endpoint the plugin preferred answers Cloudflare 403 to third-party clients | fixed |
| Riwyat | `$('.text-left') \|\| …` never reached its later selectors (a cheerio selection is always truthy), and the theme's invisible decoy blocks were returned as text | fixed |
| Kol Novel | the body regex required a `bottomnav` terminator the theme dropped, and hard-coded a `<div>` wrapper where the theme serves `<article>` | fixed — proxy-verified, not end-to-end |
| Free Kol Novel | the same regex break, plus chapter links point at `kolnovel.com` so the plugin requested `free.kolnovel.com/https://kolnovel.com/…` | fixed — proxy-verified, not end-to-end |
| Markazriwayat | not reproduced: the site serves a "Coming Soon" page and 404s every other path | not verified — no code changed |

Folded in as well: Riwyat's novel metadata on the site's new theme (name, cover, summary, author, genres, status and rating all came back empty) through additive fallbacks that only run after the existing selectors yield nothing, so classic-theme sources cannot change behaviour. Issue #2568, a kolnovel empty chapter reported after this PR opened, is the same template defect and is covered by the same fix.

## Verification

- Galaxy Novels: `npm run check:plugin` — popular 20 ✅, search 1 ✅, parseNovel 30 chapters ✅, parseChapter 13258 chars ✅ (was a 404).
- Riwyat: popular 12 ✅, search 1 ✅, parseNovel 15 chapters ✅, parseChapter 16910 chars ✅ (was 0 chars).
- The shared `lightnovelwp` fix was independently reproduced on a reachable source, TC & Sega: `parseChapter` FAIL (0 chars) → PASS 5905 chars.
[...]

Issue #2522, opened 2026-09-14
[kolnovel] Empty chapter: the dominating — العودة بنجاح


mozart2030:
### Plugin

‎العربية: Galaxy Novels, ‎العربية: مركز الروايات

### Severity

Can't Load Novels

### Steps To Reproduce

hi, there is many arabic  plugins not working as when you open a chapter it is show a massage telling you no content 
1.kol novel
2.free kolnovel 
3.riwyat ........est

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/75973fe7-ed88-4f06-a1f2-0dc094fbcd59" />

### Expected Behavior

the chapter is readable 

### Actual Behavior

the content of the chapter not loading but the cover and the novels discretion numbers and names of chapters doses load normally 

### LNReader Version

2.0.3

### Plugin Version

1.1.0

### Android Version

أندرويد 12

### Acknowledgements

- [x] I have searched existing issues and confirmed this is not a duplicate.
- [x] I have written a clear and descriptive title.
- [x] I have updated the app to the latest version **[2.0.0](https://github.com/LNReader/lnreader/releases/tag/v2.0.0-beta.3)**.
- [x] I have updated all installed plugins to the latest version.
- [x] I understand that app-specific issues should be reported in the [app repository](https://github.com/LNReader/lnreader/issues/new/choose).

PR #2530, opened 2026-09-16
chore(arabic/markazriwayat): park broken source and point to Galaxy Novels

RibatTRW:
## Change

Disposition of the fifth source left unverified by #2529. Related to #2522 — deliberately no closing keyword, since this removes rather than repairs: `markazriwayat.com` serves no novel content on any route. Direct requests hit a Cloudflare challenge, behind which sits only a Hostinger "Coming Soon" placeholder and a theme 404 page. There is nothing left to parse.

- `plugins/arabic/Markazriwayat.ts` → `plugins/arabic/Markazriwayat.broken.ts`, the repo's documented convention for an unavailable source (excluded from production compilation, source retained). The header comment names the successor: the operators' own Coming Soon page announces Galaxy Novels (`galaxynovels.com`) as their affiliated site, already shipped here as the `galaxynovels` plugin.
- Follow-up commit on this branch: the live checker skips parked `*.broken.ts` sources (workflow target exclusion plus an INCONCLUSIVE short-circuit in `scripts/live-check-plugin.js`, documented in `docs/testing.md`), since checking them can only fail by design.
- Not added to `BLACKLIST.json` (owner-takedown entries, not downtime). No version bump, per the #1977 retirement precedent.

## Verification

`build:compile` exits 0 with no `*markaz*` in output; the manifest reports the source broken and carries zero `markazriwayat` ids. Lint (2 errors / 10 warnings) and format findings are identical before and after — all pre-existing in untouched files. No live check is possible: the site serves no content, which is the reason for the PR.

Left untouched deliberately: generated `keys.json` / issue-template keys (rewritten by workflow after release) and the icon asset, matching other parked plugins. AI-authored change, no human review; the successor detail was confirmed via a second egress route.


RibatTRW:
Pushed d35f06ab for this head: the live check now skips a parked `*.broken.ts` source (INCONCLUSIVE) instead of failing it on its dead site, while healthy plugins are still fully checked. Local lint, format and compile pass; the renamed source is excluded from the build. The check run for this head is waiting on approval — could you approve it so the check can report?

